PR #2, opened 2026-01-27
Created src ixa skeleton

KOVALW:
This pull request introduces the foundational structure for running a simulation model with parameter handling and model execution logic. It adds new modules for model execution and parameter management, integrates them into the main application entry point, and establishes a mechanism for validating and retrieving global simulation parameters. Required for running the Cargo Build & Test workflow

* Added new modules `parameters` and `model` to the codebase, and re-exported their key interfaces (`ParametersContextExt`, `Params`, and `run_model`) in `src/lib.rs` for use in other parts of the application.
* Updated `src/main.rs` to initialize the simulation context, retrieve parameters, run the model with the specified seed and maximum time, and handle profiling output.
* Implemented the `run_model` function in `src/model.rs`, which initializes the random number generator, schedules a shutdown at the specified maximum simulation time, and integrates with the `ixa` framework.
* Added the `Params` struct in `src/parameters.rs` to encapsulate simulation parameters, implemented input validation for `max_time`, and provided a trait (`ParametersContextExt`) for convenient parameter access within the simulation context.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -1,6 +1,6 @@
 use ixa::{Context, ContextRandomExt, ExecutionPhase, IxaError};
 
-pub fn run_model(context: &mut Context, seed: u64, max_time: f64) -> Result<(), IxaError> {
+pub fn initialize_model(context: &mut Context, seed: u64, max_time: f64) -> Result<(), IxaError> {
     // Initialize the random number generator with the provided seed
     context.init_random(seed);
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,6 @@
 pub mod model;
 pub mod parameters;
 
-pub use model::run_model;
+pub use model::initialize_model;
 pub use parameters::ParametersContextExt;
 pub use parameters::Params;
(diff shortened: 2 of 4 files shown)


RobertJacobsonCDC (approved):
Seems reasonable to me. Some minor, mostly editorial comments:

- You do some parameter validation in `run_model` and some in `validate_inputs`. That's not necessarily a bad thing. I would suggest that parameters that the model itself cares about should be validated in `run_model`, while additional parameters (e.g. output paths) unrelated to how the model runs can be validated elsewhere. On the other hand, maybe the only parameters worth validating are all model related, in which case why not call `validate_inputs` from within `run_model`? Another way to do it is to have `run_model` just assume all parameters are valid and put all validation into `validate_inputs` so validation logic isn't spread across the codebase.
- This is just a skeleton, of course, but I don't think you ever call `validate_inputs`.
- `run_model` doesn't actually start the event loop. Maybe give it a more descriptive name?

src/model.rs
confunguido: What's the reason to introduce a run_model function? Will it take eventually all the parameters as arguments to initialize other modules or will it read the parameters directly?

confunguido (approved):
I made one comment that I would like to clarify before merging. Thanks!

PR #8, opened 2026-01-30
Importation data generating scripts

KOVALW:
This pull request introduces a new `importation` package for generating input files that determine case importation rates, along with its supporting infrastructure, documentation, and tests. The package includes ETL and data generation utilities, is fully documented, and is integrated into the CI pipeline. It also transitions the project to use `uv` for Python dependency management.

In order to generate the example documentation, call `make uv-sync` and then build the document using

```
uv run quarto render packages/importation/docs/importation_model_examples.qmd
```

**New importation package:**

* Added new package `importation` with ETL functions (`etl.py`) for retrieving and processing COVID-19 case data, and data generation utilities (`generate_outputs.py`) for creating synthetic datasets and sampling undetected infections.
* Added package metadata (`pyproject.toml`) specifying dependencies (`numpy`, `polars`, etc.) and build system.
* Added package documentation (`README.md`) describing package purpose and main functions.

**Repo updates**
* Added hooks to connect Rust `src` code with the python workspaces as well as a `__main__.py` script to be called for executing the pre run and model run steps of generating the model output. 
* Added comprehensive unit tests for ETL and data generation functions to ensure reliability and correctness (`tests/test_etl.py`, `tests/test_generate_outputs.py`). 
* Added a GitHub Actions workflow to automatically run Python tests on pull requests and pushes to `main` (`.github/workflows/run-pytest.yaml`).
* Updated project documentation (`README.md`) to instruct users to use `uv` for Python dependency management and initialization.
* Added `pytest` to the main dependencies in `pyproject.toml` for consistent testing environment.

### DIFF
--- a/packages/importation/src/importation/geographies.py
+++ b/packages/importation/src/importation/geographies.py
@@ -1,16 +1,21 @@
-from us import states
+import os
+from pathlib import Path
+
+import polars as pl
 from census import Census
 from dotenv import load_dotenv
-import os
-import polars as pl
-from pathlib import Path
+from us import states
+
 
 def get_api_key() -> str:
     """Get the Census API key from the environment variable"""
     load_dotenv()
     return os.getenv("CENSUS_API_KEY")
 
-def get_total_state_population_data(year: int | None = None, cache: bool = True) -> pl.DataFrame:
+
+def get_total_state_population_data(
+    year: int | None = None, cache: bool = True
+) -> pl.DataFrame:
     """
     Retrieve state population data from the US Census API.
 
@@ -19,9 +24,9 @@ def get_total_state_population_data(year: int | None = None, cache: bool = True)
     Otherwise, the data will be fetched from the API and optionally cached for future use.
 
     Args:
-        year (int | None, optional): The year for which to retrieve population data. 
+        year (int | None, optional): The year for which to retrieve population data.
             If None, the latest available data will be fetched. Defaults to None.
-        cache (bool, optional): Whether to cache the retrieved data locally. 
+        cache (bool, optional): Whether to cache the retrieved data locally.
             Defaults to True.
 
     Returns:
@@ -36,7 +41,7 @@ def get_total_state_population_data(year: int | None = None, cache: bool = True)
 
     Notes:
         - The function uses the Census API and requires a valid API key.
-        - Cached files are stored in the ".cache" directory with filenames in the format 
+        - Cached files are stored in the ".cache" directory with filenames in the format
           "state_population_data_<year>.csv".
     """
     filename = f"state_population_data_{year if year else 'latest'}.csv"
@@ -46,7 +51,9 @@ def get_total_state_population_data(year: int | None = None, cache: bool = True)
     else:
         api_key = get_api_key()
         c = Census(api_key, year=year)
-        state_population_data = c.acs5.state(("B01003_001E", "NAME"), Census.ALL)
+        state_population_data = c.acs5.state(
+            ("B01003_001E", "NAME"), Census.ALL
+        )
         state_population_df = pl.DataFrame(state_population_data).select(
             [
                 pl.col("NAME").alias("state_name"),
@@ -60,13 +67,16 @@ def get_total_state_population_data(year: int | None = None, cache: bool = True)
             state_population_df.write_csv(filepath)
     return state_population_df
 
-def get_state_proportion_population_data(state: str, year: int | None = None, cache: bool = True) -> float:
+
+def get_state_proportion_population_data(
+    state: str, year: int | None = None, cache: bool = True
+) -> float:
     """
     Get the population of a specific state as a proportion of the total US population.
 
     Args:
         state (str): The name, FIPS code, or abbreviation of the state.
-        year (int | None, optional): The year for which to retrieve population data. 
+        year (int | None, optional): The year for which to retrieve population data.
             If None, the most recent data is used. Defaults to None.
         cache (bool, optional): Whether to use cached data if available. Defaults to True.
 
@@ -80,7 +90,11 @@ def get_state_proportion_population_data(state: str, year: int | None = None, ca
     if state is None:
         raise ValueError(f"Invalid state name, fips, or abbreviation: {state}")
     state_name = state.name
-    state_population_df = get_total_state_population_data(year=year, cache=cache)
+    state_population_df = get_total_state_population_data(
+        year=year, cache=cache
+    )
     total_population = state_population_df["population"].sum()
-    state_population = state_population_df.filter(pl.col("state_name") == state_name)["population"].item()
+    state_population = state_population_df.filter(
+        pl.col("state_name") == state_name
+    )["population"].item()
     return state_population / total_population
--- a/packages/importation/tests/test_geographies.py
+++ b/packages/importation/tests/test_geographies.py
@@ -1,10 +1,18 @@
 from importation.geographies import get_state_proportion_population_data
 
+
 def test_get_state_proportion_population_data():
     # Test that the function returns a DataFrame with the expected columns and values
     state = "Alabama"
     year = 2020
-    result = get_state_proportion_population_data(state=state, year=year, cache=False)
-    
+    result = get_state_proportion_population_data(
+        state=state, year=year, cache=False
+    )
+
     for entry in ["Alabama", "alabama", "01", "AL", "al"]:
-        assert get_state_proportion_population_data(state=entry, year=year, cache=False) == result
+        assert (
+            get_state_proportion_population_data(
+                state=entry, year=year, cache=False
+            )
+            == result
+        )


scripts/generate_indiana_importation.py
```diff
@@ -0,0 +1,34 @@
...
+
+    data = importation.get_data()
+    indiana_importation = importation.create_proportional_state_imports(
+        data=data,
+        replicates=replicates,
+        max_infections=max_infections,
+        seed=seed,
+        state_fraction=indiana_relative_importation,
+    )
+    indiana_importation.write_daily_case_counts("./input/indiana_data.csv")
```
confunguido: I see a couple of issues with this approach. One about ixa, the other about calibration. 

I think there's an important distinction here to make about what goes into ixa and what's the outcome of this process. I think ixa should only ever need to read one replicate of these importations. Different simulations of ixa could read different replicates but that's just an input file. You could instead do something like: 

1) You save these N replicates (e.g., 500) as N different files somewhere and every time you run ixa, you indicate which file in the json config file (e.g., `importations_file = "importations_indiana_1.csv"). 

2)  You save the file as is, but you need a pre-processing script that would generate one file for x replicate. 

A different question here is how you plan to incorporate this into the calibration process. Are you expecting to always sample from these importation files? My expectation was that we would be able to calibrate the parameters that generate these curves as well as the transmission parameters.

confunguido (changes requested):
I think we need to first resolve some relevant questions before moving forward.

packages/importation/src/importation/generate_us_imports.py
```diff
@@ -0,0 +1,295 @@
...
+                + pl.col("prop_detected_symptomatic")
+                + pl.col("prop_deaths")
+            ).alias("total")
+        )
+        .filter((pl.col("total") - 1.0).abs() > 1e-10)
+        .height
+        == 0
+    ), "Proportions do not sum to 1"
+
+    return prop_ascf
```
erik-rosenstrom: We need to translate key parameters from this dataframe into input.json at some point no? We probably need to wait until we get asymptomatic individuals and deaths into the model I suppose.

scripts/generate_indiana_importation.py
```diff
@@ -0,0 +1,34 @@
+import argparse
+
+import importation
+
+
+def main(plot: bool):
```
erik-rosenstrom: Are we planning to have a repo level makefile to run this script in combination with the model?

scripts/generate_indiana_importation.py
```diff
@@ -0,0 +1,34 @@
+import argparse
+
+import importation
+
+
```
erik-rosenstrom: While this importation process doesn't seem too computationally intensive, I could see others like running the full stan model for time-varying infectiousness taking more time. Does this motivate an approach where we run a pre-processing step which generates multiple replications of output for a model, then attach each replication to an ixa model replication? The assumption being that it is more cost effective from a computation perspective to run say 500 replications a single time than one replication 500 times across individual serial tasks.

erik-rosenstrom (changes requested):
I think the main change is around replicates but overall I think this is really good stuff.

scripts/generate_indiana_importation.py
KOVALW: I think we ultimately want something that we can sample from and we would store that object in a further up preprocessing step. Similar to how the Stan model generated during the individual effects of isolation guidance project was created and then  sampled from, which takes no time. We can move forward on how to store these objects in the future

packages/importation/src/importation/model.py
```diff
@@ -0,0 +1,164 @@
...
+                    with_replacement=False,
+                    seed=seed,
+                )
+            case _:
+                raise ValueError(
+                    f"Unknown state model type: {self.model_type}"
+                )
+
+
+class ImportationModel:
```
confunguido: I'm not sure why you need three classes for this. Each of those classes only have one method. Can we combine them all into the `ImportationModel` class?

packages/importation/src/importation/etl.py
```diff
@@ -0,0 +1,119 @@
...
+    params_file = "sensitivity/covid_params_estimate.csv"
+
+    # Get default scenario parameter estimates
+    data = read_bytes(url, params_file)
+    params = pl.read_csv(data).filter(pl.col("Scenario") == "Default")
+
+    # In lieu of posterior rho_travel estimates, use beta distribution parameters
+    # E[rho_travel] = 0.5, SD[rho_travel] \approx 0.2
+    params = params.with_columns(
+        pl.lit(3).alias("rho_travel_alpha"), pl.lit(3).alias("rho_travel_beta")
```
confunguido: why 3?

packages/importation/src/importation/model.py
KOVALW: This is to keep things modular and leave open new methods for national and state models. I think it doesn't hurt to keep them as separate classes

Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: all uv-sync maturin-dev epicovid
+all: uv-sync maturin-dev epicovid
+
+uv-sync:
+	uv sync --all-packages --all-extras --dev --locked
+maturin-dev:
+	uv run maturin develop
+epicovid: uv-sync maturin-dev
```
erik-rosenstrom: I know Guido suggest removing the connect between ixa-epi-covid and the importations but it is taking a long time to build pyepicovid when I make the MakeFile

packages/importation/src/importation/model.py
```diff
@@ -0,0 +1,164 @@
...
+                from .perkins_et_al_methods import validate_data
+
+                validate_data(data)
+                self.data = data
+            case _:
+                raise ValueError(
+                    f"Unknown national model type: {self.national_model_type}"
+                )
+
+    def sample_national_importation_incidence(
```
erik-rosenstrom: I think this should be deleted for now since it doesn't appear to be used

packages/importation/docs/example_indiana.qmd
```diff
@@ -0,0 +1,113 @@
...
+)
+
+```
+
+This model assumes that the importation into a state is proportional to the US importation, and that we are using the multinomial model of undetected infections from *Perkins et al. 2020* to determine the importation of all infections into the US. The data is automatically validated for the national model method selected by the user.
+
+In order to sample the importation of infections into the state of Indiana, we use the class function `sample_state_importation_incidence`, which first calls a sample of the national model given the supplied data and then uses that national importation curve to sample the state importation curve. The function returns a Polars DataFrame with the daily number of imported infections into Indiana for each replicate.
+```{python}
+imported_infections_indiana = (
+    importation_model.sample_state_importation_incidence(state="IN")
```
erik-rosenstrom: One thing I am hesitant about with the structure of this approach is that I am not sure it would be sound to use our proposed importation method on every state. For example, NY would be very different because we know about actual international imports into the state. It makes me think the state should be an input of the `ImportationModel`

packages/importation/docs/example_indiana.qmd
```diff
@@ -0,0 +1,183 @@
...
+    all_importation_curves.append(
+        replicate_data.with_columns(
+            pl.lit(rep).alias("replicate")
+        )
+    )
+
+all_importation_curves_df = pl.concat(all_importation_curves)
+```
+
+And then we can plot to see the variance in importation curves associated with a particular set of parameters.
```
eqmooring: Maybe it's just me, but the word "variance" makes me think specifically of sigma^2; to me "variability" is more what we're getting at.

```suggestion
And then we can plot to see the variability in importation curves associated with a single set of parameters. (This plot reflects only variability that arises from sampling, not any parameter uncertainty.)
```

packages/importation/docs/example_indiana.qmd
eqmooring: Should there be a README for preparing the documentation, something that includes the information from the PR about 

`make uv-sync`
`uv run quarto render packages/importation/docs/example_indiana.qmd`

Also, it appears that `quarto` (and then `tinytex`) need to be added manually. I'm not sure if there's a way to cover them as part of formal dependency management, or if this should just be addressed in the overall "README" for the repo.

packages/importation/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+- ETL functions to process raw case data into a format suitable for modeling.
+- Functions to generate synthetic datasets of undetected infections based on observed cases and deaths.
+
+## Background
+This package depends on collecting data from *Perkins et al. Estimating unobserved SARS-CoV-2 infection in the United States. PNAS (2020).*
+The raw linelist data and importation data are available through the associated code repository of the paper, which can be used
+to re-create the national US importation model in the paper. From there, state-level importation models can be developed.
+
+## Getting started
```
eqmooring: It appears that `quarto` (and then `tinytex`) need to be added manually. I'm not sure if there's a way to handle them as part of formal dependency management, or if this fact should just be addressed in this "README" or perhaps in the overall "README" for the repo (since this likely won't be the only place we'll use Quarto for documentation).

packages/importation/src/importation/perkins_et_al_methods.py
```diff
@@ -0,0 +1,321 @@
...
+                              Defaults to None.
+    Returns:
+        pl.DataFrame: A DataFrame containing the following columns:
+            - replicate: The replicate index.
+            - prop_asymp: Proportion of asymptomatic cases.
+            - cfr: Case fatality risk.
+            - rho_travel: Reporting probability for imported symptomatic infections.
+            - prop_undetected_symptomatic: Proportion of undetected symptomatic cases.
+            - prop_detected_symptomatic: Proportion of detected symptomatic cases.
+            - prop_deaths: Proportion of deaths.
```
eqmooring:
```suggestion
            - prop_deaths: Proportion of all infections the lead to death.
```

eqmooring (approved):
Overall, I think this looks good. I added various comments, often related to just making sure that we are clear about where parameters come from.

I think that this approach differs from Perkins et al. in a number of ways, but that's okay. I don't think we have to be tied to their approach. We might want to say that we implement an approach "informed" or "motivated" or "inspired" by Perkins et al., so it's clearer that we're not fully replicating everything in their analysis.

Here is my synopsis of how the algorithm here works:

- Goal is to get a time series of the count of imported infections per day (regardless of detection) into a jurisdiction (e.g., Indiana)
- The input data are *exclusively* detected international imported cases (some of which lead to death). Data on deaths (or cases, for that matter) stemming from domestic transmission are *not* used in the `importation` package.
- Assumptions are made (potentially based on posteriors from Perkins et al) about parameters: probability that infections are symptomatic, probability that imported symptomatic cases are detected, and the probability that symptomatic cases die (and therefore are detected).
- Using these parameters and conditioning on the cumulative number of detected imported cases and imported deaths (which are by definition detected), it is possible to calculate the probability of various possible values of the total number of importations.
- Sampling proportional to these probabilities gives us a distribution of how many importations occurred.
- Once we have estimates of the cumulative number of importations, we need to attribute these importations to particular points in time. This is done by sampling from the observed distribution of onset dates of detected imported cases.
- To get importations into Indiana specifically, we sample from the US-level time series of international importations with probability equal to Indiana's proportion of the US population.

There are some limitations to this approach:
- It does not model within-US transmission and the implications for domestic inter-state importation.
- It assumes that the probability of detection (even of international importations) is constant through time.

If there's shared agreement about my synopsis and limitations listed above, I think it'd be good to add some of this to the QMD for documentation.

packages/importation/docs/importation_model_examples.qmd
```diff
@@ -0,0 +1,214 @@
...
+
+The `importation` module provides tools to fetch data, create proportional state imports, and visualize cumulative infection importation counts.
+
+## Sampling importation data
+
+In order to sample the importation of infections into the state of Indiana, we use the class function `sample_state_importation_incidence`, which first calls a sample of the national model given the supplied data and then uses that national importation curve to sample the state importation curve. The function returns a Polars DataFrame with the daily number of imported infections into Indiana for each call of the function. For the `'proportional'` method, a proportion parameter must be specified.
+```{python}
+indiana_population_proportion = 0.023
+imported_infections_indiana = (
+    importation_model.sample_state_importation_incidence(
```
confunguido: Make sure time (day) is f64

PR #23, opened 2026-02-18
Read importation file to model

KOVALW:
Support for importing infection cases from a CSV file, in addition to the existing initial infection seeding mechanism.

**Feature: Importing infections from a file**
- Added a new `import_cases_from_file` configuration option to `input/input.json` and the `Params` struct, allowing users to specify a CSV file with timed imported infections. 
- Implemented the `ImportCasesFromFile` struct and related logic in `infection_propagation_loop.rs` to read the CSV and schedule infections at specified times using the new `plan_n_infections` helper function. 
- Added deserialization for importation records and error handling for missing filenames when import is enabled.
- Added a sample CSV file `input/importation_test.csv` for testing the importation feature, containing time-stamped imported infection counts.

**Out of scope**
- Rearrange testing suite for cleaner integration tests
- Refactor infection propagation loop to be separate initialization and main loop of infection propagation
- Script to generate imported infections, as this will inherently require making choices on python - rust - model runner API

### DIFF
--- a/src/infection_importation.rs
+++ b/src/infection_importation.rs
@@ -25,8 +25,11 @@ pub struct ImportationRecord {
 }
 
 trait NovelInfectionContextExt: PluginContext + ContextEntitiesExt + InfectiousnessRateExt {
+    /// Seeds an infection for a given target person by sampling an infection time from a uniform distribution over the infectious period and then planning an infection for that person at the sampled time.
+    /// The method samples the infection time in this way to ensure that the full infectious period is not elapsed by time t=0.0
+    /// From the sample method reliance on `.get_person_rate_fn()`, the rate functions must already be loaded into the model when an infection is seeded.
     fn seed_infection(&mut self, target_id: PersonId) {
-        // Sample offset for the infectious period
+        // Sample offset for the infectious period.
         let uniform = Uniform::new(
             -self.get_person_rate_fn(target_id).infection_duration(),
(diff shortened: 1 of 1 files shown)


src/infection_propagation_loop.rs
```diff
@@ -41,6 +43,26 @@ fn schedule_recovery(context: &mut Context, person: PersonId) {
     });
 }
 
+/// Infects `infection_count` people at `infection_time` time. This is used for both seeding initial infections and importing infections from a file
+fn plan_n_infections(context: &mut Context, infection_count: usize, infection_time: f64) {
+    if infection_count > 0 {
+        context.add_plan(infection_time, move |context| {
```
erik-rosenstrom: I think it would be worth double checking that `sample_entities` won't break if there are not enough susceptible people to query.

input/importation_test.csv
```diff
@@ -0,0 +1,73 @@
+time,imported_infections
+0.0,0
+1.0,0
```
confunguido: A comment about the importations package. I still think we should generate these files without zeros, if possible. Although it doesn't affect anything in the model, it just seems unnecessary. This doesn't affect how we should read the files in ixa.

confunguido (changes requested):
I think we are almost on the same page but there are a couple of comments I would like you to consider.

PR #28, opened 2026-03-03
Documentation for current model features

KOVALW:
This PR provides minimal documentation for the current features present in the model. It would be nice to find a way to automatically enforce adding changes to the documentations when adjusting `src` code. For now, the docs focus on model implementation rather than active description of all parameters or how to run the model, relying on the README to fill that role.


docs/parameters.md
```diff
@@ -0,0 +1,20 @@
+# Parameters used in the model
+Some parameters are used for generating the importation data while others are used to execute the ixa transmission model.
+
+## Importation model parameters
+The following parameters are necessary and sufficient for generating importation data using the `multinomial` methodology
+- `symptomatic_reporting_prob`: The probability that individuals who are symptomatic are detected and reported as being infected with COVID-19
+- `case_fatality_ratio`: The probability that individuals who are symptomatic will ultimately die of COVID-19
+- `proportion_asymptomatic`: The proportion of individuals infected with COVID-19 who will never develop symptoms.
+
+Generating importation data also depedns on user specification of importation model type at the national and state level.
```
erik-rosenstrom:
```suggestion
Generating importation data also depends on user specification of importation model type at the national and state level.
```

docs/transmission.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Settings, contacts, and transmission
+As discussed in [settings documentation](settings.md), settings can reflect density dependent transmission by setting category specific parameters $\alpha$. The rate of infection attempts in a setting therefore scalaes with the number of active individuals in that setting $N$ with the form $(N-1)^\alpha$. For an individual's active itinerary, the weighted average of the density dependent scaling factors is calculated from the proportion of time the individual spends in each setting. The largest setting specific modifier is tracked for each individual and used to calculate the maximum rate of infection attempts, akin to assuming that the maximum rate is that which would occur if an inidivudal spent all of their time in the setting with the highest scaled rate.
```
erik-rosenstrom:
```suggestion
As discussed in [settings documentation](settings.md), settings can reflect density dependent transmission by setting category specific parameters $\alpha$. The rate of infection attempts in a setting therefore scalaes with the number of active individuals in that setting $N$ with the form $(N-1)^\alpha$. For an individual's active itinerary, the weighted average of the density dependent scaling factors is calculated from the proportion of time the individual spends in each setting. The largest setting specific modifier is tracked for each individual and used to calculate the maximum rate of infection attempts, akin to assuming that the maximum rate is that which would occur if an individual spent all of their time in the setting with the highest scaled rate.
```

docs/initialization.md
```diff
@@ -0,0 +1,7 @@
+# Simulation Initialization
+
+## Seeding Initial Conditions
+When the simulation is instantiated, all individuals are created in the susceptible compartment. The model initializes infectious individuals partway through their infection at time 0. This is implemented using negative simulation time feature of `ixa`. Transmission is only enabled at time 0.0, so any infectious individuals seeded prior to that time are prevalent. The infectious seeding process samples from the susceptible population uising a binomial distribution with success probability of `initial_prevalence`, such that the actual initial prevalence of infectious individuals at time 0.0 is sampled with some noise around the provided parameter.
```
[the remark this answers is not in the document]
eqmooring: I think some added language in `initialization.md` and possibly a bit in `importation.md` are in order to flesh out how these processes are similar / different. For example, it is stated that initialization uses the special negative time feature, but it's not clear whether importation can / does.

docs/transmission.md
eqmooring:
```suggestion
As discussed in [settings documentation](settings.md), settings can reflect density dependent transmission by setting category specific parameters $\alpha$. The rate of infection attempts in a setting therefore scales with the number of active individuals in that setting $N$ with the form $(N-1)^\alpha$. For an individual's active itinerary, the weighted average of the density dependent scaling factors is calculated from the proportion of time the individual spends in each setting. The largest setting specific modifier is tracked for each individual and used to calculate the maximum rate of infection attempts, akin to assuming that the maximum rate is that which would occur if an inidivudal spent all of their time in the setting with the highest scaled rate.
```

PR #52, opened 2026-03-12
Updating to ixa v2.0.0-beta2

erik-rosenstrom:
This is breaking update. It requires migrating from `IxaError::IxaErrors` to a model specific error type. A new `ModelError::ModelError` created using `thiserror` replaces the previous implementation of `IxaError`. Several other error types are created to handle errors `from` other libraries. This includes a `ModelError::IxaError` for errors generated from ixa itself.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -7,7 +7,12 @@ use crate::{
     reports, settings, symptom_status_manager,
 };
 
-pub fn initialize_model(context: &mut Context, seed: u64, max_time: f64, synth_population_override: Option<PathBuf>) -> Result<(), ModelError> {
+pub fn initialize_model(
+    context: &mut Context,
+    seed: u64,
+    max_time: f64,
+    synth_population_override: Option<PathBuf>,
+) -> Result<(), ModelError> {
     // Initialize the random number generator with the provided seed
     context.init_random(seed);
 
--- a/src/parameters.rs
+++ b/src/parameters.rs
@@ -385,7 +385,7 @@ mod tests {
             ),
             ..Default::default()
         };
-        let e = validate_inputs(&parameters).err();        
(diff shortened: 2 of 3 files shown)


RobertJacobsonCDC:
I've collected some error design principles that I copied and pasted below if you are interested. Not all of it is directly applicable for this use case. Feel free to disagree with me. Error handling is kind of a black art. There are lots of philosophies and opinions about the best way to do it.

## Principles

### Prefer useful granularity

- Prefer finer-grained variants over coarse buckets when the added distinction is useful for diagnosis, debugging, or user action.
- Choose granularity based on the API boundary where the error is produced/consumed.
- Keep deeper internal detail in `Error::source()` chains when that is the right layering.

### Delineate user-facing vs internal errors

- Separate errors intended for end-user consumption from errors intended for internal handling or developer diagnosis.
- Bridge layers explicitly via `Error::source()` where appropriate.

### Preserve foreign error context

- Prefer preserving foreign error types (including `std` and dependency-crate errors) using `Error::source()`.
- Exceptions to this rule are acceptable for trivial/single-cause failures with no additional useful source information, or where external behavior compatibility constrains semantics.

### Design around recovery boundaries

- Distinguish errors to support meaningful caller actions: retry, correct input, change configuration, degrade gracefully, abort, or report bug.
- Merge technical causes when caller response is the same; split contexts when response differs.

### Make invalid states unrepresentable

- Prefer type-level constraints, validated constructors, and constrained APIs over runtime misuse errors where feasible.
- Reserve runtime errors for genuinely dynamic failure.

### Treat names/messages as API contract

- Variant names and display messages should describe domain intent precisely and stably.
- Prefer domain-language semantics over implementation-location wording.

PR #54, opened 2026-03-17
symptom transition probabilities vary by age category

### LINKED ISSUE #31: Incorporating Age Groups into symptom risk
People in the model have a property `Age`. Input for symptom status progression will supply various parameters in terms of age groups e.g, 0-18, 19-64, 65+. We need to be able to easily determine which parameters to reference for an individual given their age. We also need a clear way to read in age groups, standardize them across parameters, and  pass them between modules/tests model.

### LINKED ISSUE #53: Variable is wrong in test; should defaults be changed to make this easier to find?
This should be `probability_dead_given_critical: 0.0`

The test still passes, because the default value is also 0.0.

Would it make sense to change the defaults to make this less likely in the future?

https://github.com/CDCgov/ixa-epi-covid/blob/83a88aa78683370a6bfa056d719e4dbfd388b517/src/symptom_status_manager.rs#L416

eqmooring:
## Background
For many diseases, the risk of progression to different clinical states (more severe symptoms, deaths, etc.) is age dependent. We want to reflect this phenomenon in our model by specifying transition probabilities that depend on an individual's age group. (The delay distributions are not age specific.)

## What's changed
- Changes are made to `input/input.json`, `src/reports/mod.rs`, `src/parameters.rs`, and `src/symptom_status_manager.rs` to create a 4-level age categorization.
  - The names of the levels are hard coded in an enum in `src/symptom_status_manager.rs`.
  - The parameter `symptom_age_group_thresholds` (a vector of the lower age bounds that define each group) is added.
  - The parameters `probability_severe_given_mild`, `probability_critical_given_severe`, and `probability_dead_given_critical` are changed to hashmaps of probabilities, rather than just probability values.
  - Notably, `probability_mild_given_infect` is not made age dependent.
  - The necessary changes are made to validate these parameters and to create `SymptomAgeGroup` as a derived person property.
  - The values used in `input/input.json` are based on "real" values derived from CDC COVID-19 scenarios documents.
 - Significant changes are made to the functions in `src/symptom_status_manager.rs`.
  - These changes use the correct age-specific probability when simulating which transitions to schedule.
  - An arguably bigger change is to create a `SymptomData` enum that records when a person entered each status, rather than just tracking their current status. Consequently, `plan_symptom_transition` was replaced with a simpler `draw_transition_time` function and corresponding changes were made to the `process_symptom_change_event` and `init` functions.
  - Tests were heavily revised to reflect these changes. 

## Outstanding issues
 - A less-than-ideal feature of the current implementation is that the names of the age groups are hard coded in an enum, while the thresholds that define them are an input parameter.
 - The series of tests `test_X_to_Y_symptom_change_event` used to test `process_symptom_change_event` have been subsumed into a single test `test_age_specific_symptom_trajectories`. This latter test is in many ways more comprehensive. However, it would be possible to bring back transition-specific tests. (The transition-specific tests used `add_plan_with_phase`; the revised version of `test_no_symptoms_to_mild` now shows an alternative approach that avoids `add_plan_with_phase` to check the state of the model.)


src/symptom_status_manager.rs
```diff
@@ -1,19 +1,49 @@
...
 };
 
 define_rng!(SymptomsRng);
 
-#[derive(Serialize, Deserialize, PartialEq, Debug, Copy, Clone, Eq, Hash)]
+#[derive(Serialize, Deserialize, PartialEq, Debug, Copy, Clone)]
+pub enum SymptomData {
+    NoSymptoms,
+    Mild {
+        mild_time: f64,
```
erik-rosenstrom: I am a bit confused why we need symptom data. It does not seem to be used anywhere in the code.

erik-rosenstrom (changes requested):
Looking good. I see the how the symptom data is useful for the test, and I think it will have utility in the future. I think the major comment is really about functionalization of setting symptom data properties to clean up the code, and investigating the vector approach compared to the hash map approach

input/input.json
```diff
@@ -13,37 +13,38 @@
                 "duration": 5.0
             }
         },
+        "symptom_age_group_thresholds": [0, 18, 50, 65],
```
confunguido: As I look at this more and more, I think we need to reimplement the age groups in a more flexible way. Here's what I'm thinking. 

input:
```
symptom_age_groups: [{name: AgeXtoY, min: 0, max: 10}, ...]
```

In rust, we want to do two things. 1) To validate that all groups are valid, 2) to convert those age groups into a sorted vector, `AgeGroups: Vec<AgeGroupData>` where `AgeGroupData` contains the name, min, max. Then, we should refer to these groups by their index, which could be wrapped into a struct, e.g., `AgeGroup(u64)`. AgeGroup would then be a derived property. We then also store the probabilities in a vector than can be accessed by the agegroup index, so then we can refer to the probabilities based on that index.

confunguido (changes requested):

src/symptom_status_manager.rs
```diff
@@ -48,87 +122,168 @@ impl SymptomDelayDistLogNormParams {
...
-                    mild_to_severe_delay,
-                );
+        SymptomData::Mild { .. } => {
+            if context.sample_bool(
+                SymptomsRng,
+                *probability_severe_given_mild
+                    .get(&symptom_age_group)
+                    .unwrap(),
+            ) {
+                let transition_time = draw_transition_time(context, mild_to_severe_delay);
```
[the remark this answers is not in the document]
confunguido: I think we can leave that for another PR. This is clear enough for this point.

confunguido (approved):
This is good if you fix the conflicts and tests.

erik-rosenstrom (approved):
Looks good to me!

PR #81, opened 2026-04-02
vectorized settings first pass

erik-rosenstrom:


src/population_loader.rs
```diff
@@ -27,85 +27,74 @@ impl_property!(Age, Person);
...
+
+#[derive(Debug, PartialEq, Clone, Copy, Serialize)]
+pub struct ItineraryRatios {
+    pub itinerary_ratios: [Option<f64>; SETTING_COUNT],
+}
+impl_property!(
+    ItineraryRatios,
+    Person,
+    default_const = ItineraryRatios {
+        itinerary_ratios: [None; SETTING_COUNT]
```
bbbruce:
```suggestion
        itinerary_ratios: [0; SETTING_COUNT]
```

src/population_loader.rs
```diff
@@ -27,85 +27,74 @@ impl_property!(Age, Person);
...
-#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, Copy, Hash)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Hash)]
+pub struct WorkId(pub Option<SettingId>);
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Hash)]
 pub struct CommunityId(pub Option<SettingId>);
 
-pub trait GenericSetting {
-    fn get_setting_id(&self) -> Option<SettingId>;
-}
+impl_derived_property!(HomeId, Person, [SettingIds], [], |setting_ids| HomeId(
```
bbbruce: Not sure these are that helpful??

src/settings.rs
```diff
@@ -1,120 +1,97 @@
...
+define_global_property!(SettingRatios, [f64; SETTING_COUNT]);
 
 trait ContextSettingExtPrivate: PluginContext + ContextEntitiesExt + ContextParametersExt {
-    fn get_setting_ratio(&self, setting: SettingId) -> Result<f64, ModelError> {
+    fn get_setting_ratio(&self, setting_category: SettingCategory) -> Result<f64, ModelError> {
         let ratios = self.get_global_property_value(SettingRatios).unwrap();
-        let setting_index = self
-            .get_property::<Setting, SettingPropertyIndex>(setting)
-            .0;
+        let setting_index = setting_category as usize;
```
bbbruce: Instead of converting to `usize` in a million places you could make things like `alpha` and `impl` of SettingCategory couldn't you?

src/settings.rs
```diff
@@ -163,39 +140,49 @@ pub trait ContextSettingExt:
         .unwrap();
     }
 
+    fn initialize_setting_size(&mut self) -> Result<(), ModelError> {
```
bbbruce: What is this doing?

bbbruce (commented):
Some initial comments.

src/settings.rs
erik-rosenstrom: I was having some trouble with this. I have been able to get it to work by implementing the Index trait for the enum as follows 
```rust
// Implement immutable indexing
impl<T> Index<SettingCategory> for [T; SETTING_COUNT] {
    type Output = T;
    fn index(&self, index: SettingCategory) -> &Self::Output {
        &self[index as usize]
    }
}
```

src/settings.rs
erik-rosenstrom: This method sets the number of people in each setting as property of the setting. It uses the derived setting id person properties. It is called in the population loader immediately after the population is loaded. It enables us to look up the size rather than query it during the infection loop. In the future the size could be updated when people change settings using event subscriptions.

src/population_loader.rs
```diff
@@ -27,85 +27,74 @@ impl_property!(Age, Person);
...
+    let work_string: String = String::from_utf8(person_record.workplaceId.to_owned())?;
 
     // Add person to context
     let person_id: PersonId = context.add_entity((Age(person_record.age),)).unwrap();
-    context.add_person_to_setting(
+    context.add_person_to_settings(
         person_id,
-        SettingCategory::Home,
-        SettingCode(home_id.parse()?),
+        home_string,
```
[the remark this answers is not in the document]
erik-rosenstrom: I believe we need to read them in as strings and then parse to usize. I have moved the parse to immediately when it is read and have not seen a speed up from bench-compare

erik-rosenstrom:
```
======================================================================
  main (dc2200a) vs etr_settings_vectorized
======================================================================

  Benchmark                                      Base         Current  Change
  ---                                             ---             ---  ---
  init/population/10k                       22.531 ms       19.089 ms  -15.3% faster ✓
  init/population/100k                      284.89 ms       238.53 ms  -16.3% faster ✓
  execute/population/10k                    249.43 ms       141.59 ms  -43.2% faster ✓
  execute/population/100k                    4.5226 s        2.7721 s  -38.7% faster ✓

======================================================================
```

bbbruce (approved):
LGTM - only thing is not sure need to add the `.codex` file that one or both of us may have committed by accident... maybe add to .gitignore instead? Don't know best practice.

erik-rosenstrom:
after rebasing to main
```
======================================================================
  main (dc2200a) vs etr_settings_vectorized
======================================================================

  Benchmark                                      Base         Current  Change
  ---                                             ---             ---  ---
  init/population/10k                       19.455 ms       18.459 ms  -5.1% faster ✓
  init/population/100k                      246.15 ms       222.94 ms  -9.4% faster ✓
  execute/population/10k                    209.15 ms       120.54 ms  -42.4% faster ✓
  execute/population/100k                    3.6415 s        2.2565 s  -38.0% faster ✓

======================================================================
```

PR #82, opened 2026-04-03
Parallel calibration of phase 1

### LINKED ISSUE #22: Run model projections given first death
We will use a simple proof of concept script to calibrate the model with importation and local transmission to the timing of the first reported death due to COVID-19 reported in the state of Indiana. We will condition on day zero in the model corresponding to January 1st 2020, and filter for model trajectories that have the first case reported exactly on March 16th, 2020. We will then project these model runs calibrated to data forward for a short term analysis of how many deaths can be expected in the population without interventions

Once model functionality for this phase is complete, we need to 
- [x] Document all parameters present in the model
- [ ] Select parameters for calibration
- [ ] Tune tolerance values for the delay from Jan 1 to time of first reported death
- [ ] Connect the model to calibrationtools using MRP

### LINKED ISSUE #85: Fix degenerate seed performance in importation module
In parallel, sometimes the forked processes will not use different np.seeds, and the technique being used to set the seed in the importation module is outdated. Ideally, we would use seed sequences. To reduce change overhead, we can simply generate a default_rng for making the random choice of the weighted multinomial sample

KOVALW:
Full configuration, documentation, and workflow for the Phase 1 model calibration routine, along with improvements to the synthetic population generation tooling and bug fixes in the importation package. The most significant changes include new input files and Makefile targets for calibration, a comprehensive calibration analysis report, and packaging of the synthetic population generator.

**Running an example**
To calibrate a toy example using the `dev-config.yaml` with some population size, such as 50K, run
```
make calibrate-phase-1-dev SIZE=50_000
```

If you want to run the entire routine for the same dev config, you can run
```
make plot-phase-1-projection-dev SIZE=50_000
```

To run the single step ABC algorithm used in production, use
```
make plot-phase-1-projection
```

but it may be easiest to run the calibration stage first with
```
make calibrate-phase-1
```

**Calibration workflow and configuration:**

* Added Makefile targets `calibrate-phase-1` and `calibrate-phase-1-dev` to automate building the Rust binary and running the phase 1 calibration script with production or development configs.
* Introduced configuration files for calibration: `prod-config.yaml` and `dev-config.yaml` (specifying model, priors, and parameters), and default parameters and priors in `default_params.json` and `priors.json`. [[1]](diffhunk://#diff-88850311100f2c465ea1d65e591c11b947d4deb505230ace53c32203932b4b47R1-R10) [[2]](diffhunk://#diff-49c67c263c59c90e5757b3d1580ea0b6bc2965fa741987ba8d7dae975580f0bcR1-R10) [[3]](diffhunk://#diff-964a1f858a8986f87ba25c03303a7a4fc9e82f476e9703a58b09cb91e71dd7dcR1-R74) [[4]](diffhunk://#diff-cdadfc6da3c079287f0a6336adf166facbe919c2329773697a6ae537d3b616d8R1-R32)
* Added a detailed README (`experiments/phase1/README.md`) describing how to run the calibration and render the analysis report.

**Calibration analysis and reporting:**

* Added a comprehensive Quarto report (`calibration.qmd`) for Phase 1 calibration, including code for diagnostics, posterior/prior comparison, and visualizations of model trajectories.

**Synthetic population tooling:**

* Refactored the synthetic population script into a package (`create_synthetic_population`), added a `pyproject.toml`, and exposed its API via `__init__.py`. Updated the Makefile to call the new module. [[1]](diffhunk://#diff-3430a6a3c7b34bbb1abc9b432966b4f59a3d57b8cc2a8f67e99844263b663c78R1-R14) [[2]](diffhunk://#diff-87eb4e152304b5dcbfa513235e7c9d2e985076882148e19efb7a9ae7f6600024R1-R19) [[3]](diffhunk://#diff-5b3fd13feeeec825f640cd8aaa92db6efc172a605c9627e4d2314f7cd591e151L362-R362) [[4]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L16-R16)
* Renamed the script entry point from `main` to `run` and ensured compatibility with the new packaging structure.

**Bug fixes and maintenance:**

* Fixed a bug in `prob_undetected_infections` where the probability calculation could fail if total weight was zero.
* Cleaned up unused code in `sample_undetected_infections`.
* Updated input JSON to include `aggregated_deaths_report` and `first_death_terminates_run` keys for consistency with calibration configs.
* Removed outdated README from `ixa-epi-utils`.


Makefile
```diff
@@ -57,3 +57,17 @@ bench: input/synth_pop_people_WY_10_000.csv input/synth_pop_people_WY_100_000.cs
...
+	uv run cargo build -r
+	uv run python ./scripts/phase_1_calibration.py -c ./experiments/phase1/input/prod-config.yaml -o ./experiments/phase1/calibration/output_indiana
+
+calibrate-phase-1-smc: ./experiments/phase1/input/priors.json ./experiments/phase1/input/default_params.json
+	uv run cargo build -r
+	uv run python ./scripts/phase_1_calibration.py -c ./experiments/phase1/input/prod-smc-config.yaml -o ./experiments/phase1/calibration/smc
+
+calibrate-phase-1-dev: ./experiments/phase1/input/priors.json ./experiments/phase1/input/default_params.json
+	uv run cargo build -r
+	uv run python ./scripts/phase_1_calibration.py -c ./experiments/phase1/input/dev-config.yaml -o ./experiments/phase1/calibration/dev_$(SIZE) --default-population-size-dev $(SIZE)
```
eqmooring:
```suggestion
	uv run python ./scripts/phase_1_calibration.py -c ./experiments/phase1/input/dev-config.yaml -o ./experiments/phase1/calibration/dev_$(NORMALIZED_SIZE) --default-population-size-dev $(NORMALIZED_SIZE)
```
I tried running `make calibrate-phase-1-dev SIZE=5000` and got an error about model initialization failing because of a lack of a CSV file. Indeed, per https://github.com/CDCgov/ixa-epi-covid/blob/b8184cc0fb2505ba48fd6ea74241f5f728c917a0/scripts/phase_1_calibration.py#L299, the input for `default-population-size-dev` needs to be version of the size with the underscores.

From my testing, I don't think it's essential to change the name of the calibration results directory to use `NORMALIZED_SIZE`, but it may nonetheless be a good idea for the sake of consistency.

Makefile
eqmooring, line 1: The PR mentions rules such as `plot-phase-1-projection-dev` and `plot-phase-1-projection`, which I am not seeing here.

experiments/phase1/README.md
eqmooring, line 1: I think this is fine, though I might suggest that this README talk more about using make.

experiments/phase1/README.md
```diff
@@ -0,0 +1,21 @@
+# Running the calibration routine
+In order to generate the results analysis report from the `reports/` directory, first calibrate the model by using the phase 1 calibration script. Ensure that the uv environment is synced and the rust binaries have been assembled
+```
+uv sync --all-packages
+uv run cargo build -r
+uv run python scripts/phase_1_calibration.py
```
eqmooring:
```suggestion
uv run python scripts/phase_1_calibration.py
uv run python scripts/phase_1_projection.py
```
It looks like the output from the projection step is needed to get the calibration report to work.

experiments/phase1/input/dev-config.yaml
confunguido, line 1: The `use_env_synth_pop_file` variable is confusing.

Makefile
```diff
@@ -57,3 +57,36 @@ bench: input/synth_pop_people_WY_10_000.csv input/synth_pop_people_WY_100_000.cs
...
+	uv run python ./scripts/plot_phase_1_projection.py -d smc
+
+calibrate-phase-1-dev: ./experiments/phase1/input/priors.json ./experiments/phase1/input/default_params.json
+	uv run cargo build -r
+	uv run python ./scripts/phase_1_calibration.py -c ./experiments/phase1/input/dev-config.yaml -o ./experiments/phase1/calibration/dev_$(NORMALIZED_SIZE) --default-population-size-dev $(NORMALIZED_SIZE) --max-workers $(MAX_WORKERS)
+
+projections-phase-1-dev: ./experiments/phase1/calibration/dev_$(NORMALIZED_SIZE)/results.pkl
+	uv run python ./scripts/phase_1_projection.py -d dev_$(NORMALIZED_SIZE) --max-workers $(MAX_WORKERS) -f --plot-distances
+
+plot-phase-1-projection-dev: projections-phase-1-dev
```
confunguido, line 102: We need to improve this Makefile. I would assume that plot-phase-1-projection-dev would also call calibrate-phase1-dev

src/parameters.rs
```diff
@@ -74,6 +74,10 @@ pub struct Params {
     pub incidence_report: ReportParams,
     /// Transmission report with a name required
     pub transmission_report: ReportParams,
+    /// Aggregated incident deaths report with a period and name required
+    pub aggregated_deaths_report: ReportParams,
+    /// Terminate run on first observed death
+    pub first_death_terminates_run: bool,
```
confunguido, line 80: Is this something we want to change for phase-2? For instance, do we want a threshold to terminate simulation instead of 1 death?

scripts/phase_1_calibration.py
```diff
@@ -0,0 +1,309 @@
...
+    def particles_to_params(
+        particle: Particle, reader: ParticleReader = reader
+    ):
+        particle_params = reader.read_particle(particle=particle)
+        # Make particle-specific output directory and update the output path in the parameters accordingly
+        simulations_dir = Path(
+            particle_params["config_inputs"]["output_dir"], "simulations"
+        )
+        # Count existing directories in simulations_dir
+        if not simulations_dir.exists():
```
confunguido, line 187: Since we already noticed that seeds could repeat and cause issues, this is good to have. However, I think it's a bit cryptic. I would like to create a follow up issue for this. In my opinion, it would be easier to save as:
```
output/
 - step_0
   - sim_0/
   - sim_1/
   - sim_2/
   - ...
 - step_1
   - ...
 - ... 
```
This would allow us to also keep track of rejected particles.

scripts/phase_1_calibration.py
```diff
@@ -0,0 +1,309 @@
...
+        "count": [1],
+    }
+)
+
+
+def main(
+    config_file: str | Path,
+    output_dir: str | Path,
+    max_workers: int = 10,
+    default_population_size_dev: str = "50_000",
```
confunguido: I'm not sure I like this variable. I'm not quite understanding why we can't just put the population in the config file.

experiments/phase1/input/dev-config.yaml
```diff
@@ -0,0 +1,10 @@
+exe_file: ./target/release/ixa-epi-covid
+force_overwrite: true
+state: Indiana
```
confunguido, line 3: This would give the impression that it will also select State+Year in the synthetic population, which it doesnt.

scripts/plot_phase_1_projection.py
confunguido, line 1: Why is there a plot projection but not a plot calibration script?

confunguido (approved):
I don't have any major concerns with this PR but would like to get my comments addressed in future issues.

src/abort_run.rs
```diff
@@ -0,0 +1,24 @@
...
+            move |context, event| {
+                if event.current == SymptomStatus::Dead {
+                    context.add_plan(context.get_current_time() + 1.0, move |context| {
+                        context.shutdown();
+                    });
+                }
+            },
+        );
+    }
+}
```
erik-rosenstrom, line 24: Nit: but it would be good to have a test for this

README.md
```diff
@@ -62,6 +62,34 @@ make profile PROFILE_SIZE=10m PROFILE_FEATURES=profiling  # both
...
+
+`make calibrate-phase-1-dev` is the easiest way to run the calibration routine of Phase 1, in which we compare the first observed deaths in the model to the first observed death in the State of Indiana. Running this command with setting `SIZE` will generate a toy synthetic population of that size based on Indiana PUMs census data and then run the Phase 1 calibration script. For example, to run routines in popualtions 50K, 100K, and 1 million:
+
+```bash
+make calibrate-phase-1-dev SIZE=50_000
+make calibrate-phase-1-dev SIZE=100_000
+make calibrate-phase-1-dev SIZE=1_000_000
+```
+
+By default, the calibration routine uses four parallel worker threads to run simulations. Depending on your computer specs, this can be altered by setting the `MAX_WORKERS` parameter
```
erik-rosenstrom, line 75: It might help to explain with a sentence what a worker conceptually is, and how to pick the number of workers. I.e., don't pick too many or you can run out of memory.

scripts/phase_1_calibration.py
```diff
@@ -0,0 +1,309 @@
...
+
+    # Handle synth population file options ----------------------------------------------
+    # Users may specify a SYNTH_POP_FILE of their choosing in the .env file, which will then be copied into the local repo input path for running
+    # Otherwise, the default population file is used, and can be created automatically using the create_synthetic_population package if it does not already exist at the specified path
+    load_dotenv()
+    ixa_overrides = {
+        "max_time": config.target_data["t"][0] + config.tolerance_values[0] + 1
+    }
+
+    synth_pop_file_env = os.getenv("SYNTH_POP_FILE")
```
erik-rosenstrom, line 56: The synthetic population piece seems like it could be moved to a helper method

scripts/plot_phase_1_projection.py
```diff
@@ -0,0 +1,323 @@
...
+                aggregated_deaths_report = (
+                    aggregated_deaths_report.with_columns(
+                        pl.col("count")
+                        .cum_sum()
+                        .over(order_by="t_upper")
+                        .alias("cumulative_count"),
+                        pl.lit(particle_dir.name).alias("seed"),
+                    )
+                )
+                prevalence_report = (
```
erik-rosenstrom, line 58: Maybe we should consider adding age group to reports so stuff like this is already taken care of

scripts/phase_1_projection.py
```diff
@@ -0,0 +1,260 @@
...
+            particle_params["config_inputs"]["output_dir"], "simulations"
+        )
+        # Count existing directories in simulations_dir
+        if not simulations_dir.exists():
+            dir_count = 0
+        else:
+            dir_count = len(os.walk(simulations_dir).__next__()[1])
+        output_dir = Path(
+            simulations_dir,
+            ".".join(
```
erik-rosenstrom, line 163: What's the motivation for saving the seed in the folder name? Its in the in the `input.json` within the folder

scripts/phase_1_projection.py
```diff
@@ -0,0 +1,260 @@
+import argparse
```
erik-rosenstrom, line 1: When running the dev for 50k I get multiple printouts with the following error message
```
Exception ignored while calling deallocator <function Image.__del__ at 0x729e0687e8d0>:
Traceback (most recent call last):
  File "/home/xui6/.local/share/uv/python/cpython-3.14.3-linux-x86_64-gnu/lib/python3.14/tkinter/__init__.py", line 4283, in __del__
    self.tk.call('image', 'delete', self.name)
RuntimeError: main thread is not in main loop
```

erik-rosenstrom: Also there seems to be an issue with the replication number 

Image

erik-rosenstrom (changes requested):
Looking good. I have a couple of issues running

scripts/phase_1_projection.py
KOVALW: The combination of seed an simulation id ensures uniqueness (simulation id by particle slot or by step are not guaranteed to be unique, and seeds can be repeated with probability $1-\left[\left(1- \frac{1}{\text{MAXINT}}\right)^{\text{REPLICATES}}\right]$

scripts/phase_1_projection.py
KOVALW: I currently use the count of number of folders in the directory to assign the integer. If two parallel processes begin at a similar enough time and accept a particle each, then the same integer is used, hence the jump from 0, 0, and 1 to 3

scripts/plot_phase_1_projection.py
KOVALW: Opening as an issue for phase 2 #100

Makefile
KOVALW: #101

src/parameters.rs
KOVALW: #102

scripts/phase_1_calibration.py
KOVALW: I'm really hesitant to create this quantity of output files for both computation efficiency and storage. I don't think there's a big value add beyond storing the run conditions for model runs that yield errors above the current accepted values. I think alongside run conditions, we'll store accepted particle parameter values and their slot id. Moving this discussion to cfa-calibration-tools [here](https://github.com/CDCgov/cfa-calibration-tools/issues/24)

scripts/phase_1_calibration.py
KOVALW: I think for development we should keep this on the command line, mirroring the benchmarking we do with the same parameter value

experiments/phase1/input/dev-config.yaml
KOVALW: It does do so when we call `create_synthetic_population`. I've made an issue to clarify importation vs synthetic population file values here #103

scripts/plot_phase_1_projection.py
KOVALW: Calibration plots are only generated in the report for right now (the distance error distribution and the posterior parameter value distributions) #104

src/abort_run.rs
```diff
@@ -22,3 +22,69 @@ pub fn init(context: &mut Context) {
...
+                    critical_time: 0.0,
+                    dead_time: 1.0,
+                },
+            );
+        });
+        context.add_plan(3.0, move |context| {
+            let _ = context.add_entity::<Person, _>((Age(23),)).unwrap();
+        });
+        context.execute();
+        assert_eq!(context.get_entity_count::<Person>(), 2);
```
erik-rosenstrom, line 66: Nit: It may be better to use the `context.get_current_time()` after `context.execute()` to ensure that the simulation stops at time 2, one day after the person dies

erik-rosenstrom (approved):
Looking good to me! Just one nit comment

PR #84, opened 2026-04-06
feat: FIPSCode-based synthetic population parsing, ASPR-compatible population loading and synth pop generation

RobertJacobsonCDC:
This PR adds:

1.  a 64-bit binary representation of setting IDs (setting codes, `homeId`, `schoolId`, `workplaceId`), called `FIPSCode`,  with accessors for the data they encode
2.  noncopying byte-slice oriented parsers to convert from textual setting IDs (setting codes) into `FIPSCode` values
3.  a fast file parser / reader for datasets formatted like the ASPR synthetic population dataset, including support for transparently reading from data in zip archives ~and some convenience methods for reading from the ASPR dataset specifically~ (Edit: removed ASPR-specific stuff)
4.  wiring of model code (`population_loader`) to use the new `FIPSCode` type for `SettingCode` and to use the new file reader facilities to load synth pop data
5.  modifications to `scripts/create_synthetic_population.py` to generate setting IDs in the same format as the ASPR dataset so we can use the same parser / file reader code for both

The code for (1)-(3) takes the form of a new module `pop_reader` that includes code originally in the `ixa-fips` crate and the defunct `ixa-aspr` crate. The plan is to incubate this code in `ixa-epi-covid` and then backport it to `ixa-fips` when we are confident it has general utility. 

**Out of scope for this PR:**

- Changing any model code whatsoever except to swap out the synthetic population file reader and to swap the definition `SettingCode` (previously a `usize` newtype) with `FIPSCode` (now a type alias `SettingCode = FIPSCode`).
  - no optimizations leveraging the fact that `SettingCode` / `FIPSCode` now carries its setting category
- Changes to how `scripts/create_synthetic_population.py` generates synthetic population data is limited only to the "sequential ID" suffix.
  - the sequential ID was changed from a global counter to a "within-tract" counter to align with the ASPR dataset
  - the formatting of the suffix was aligned with the format used in the ASPR dataset, including allowing a potential extra digit for overflowing sequential IDs
  - the data sources and other mechanisms for synthesis were left unmodified
  - no distinction between public & private schools was added
- Choices of data sources and additional validation of input data. If it parses to a `FIPSCode`, it's usable.

**Housekeeping work done:**

- Removed several unused crate dependencies
- Added `scripts/syn-pop-generation.md` documenting how `scripts/create_synthetic_population.py` works.
- Added `test-syn-pop` command to `Makefile` that runs `uv run pytest scripts/test_create_synthetic_population.py`.
- Added `clean-synthetic-population` command to `Makefile` that takes optional STATE and SIZE parameters that removes _generated_ `input/synth_pop_people_*` and `input/synth_pop_region_*` files.
- gitignore additions for IDE and macOS junk.

## Modularity of FIPS Code / Population Reader Functionality

Even though we've brought the code under a single module while it is being incubated, we still organize the FIPS Code representation, parser functionality, and file reader utilities in distinct layers, the deeper layers depending on the earlier layers, so that when we return it to a separate crate users can choose to pull into their projects only what they use.

### Layer 1: Compact, efficient FIPS Code binary representation

Previously in its own crate `ixa-fips` but now incubated in the `pop_reader::fips_code` module, we have:

- **`FIPSCode`:** a 64-bit compact binary representation of FIPS geographic region codes (and “code fragments”)
  - Accessor methods for component "fields"
  - Convenience constructor methods
- **`ExpandedFIPSCode`:** an "expanded" struct representation in which the `FIPSCode` fields are real, explicit fields. You can easily convert between `FIPSCode` and `ExpandedFIPSCode`.
- **`USState`:** An enum for U. S. states, because *numeric state code* to state conversion is not a simple linear map.

This layer is concerned only with binary representation. It is intended to be generally useful to represent things having a FIPS hierarchical geo ID tract prefix, but our motivating use case is to represent `homeId`, `schoolId`, and `workplaceId` values as they are used, for example, in the ASPR dataset.

### Layer 2: Setting ID / FIPS Code Parsing

Previously in its own crate `ixa-aspr` but now incubated in the `pop_reader::parser` module, we have utilities for parsing setting codes as they appear in the ASPR synthetic population data set:

- **High-level parsers** of setting code strings to `FIPSCode`: `parse_fips_home_id`, `parse_fips_school_id`, `parse_fips_workplace_id`.
- **Low-lever component parsers:**  `parse_county_code`, `parse_tract_code`, etc.
- **`PopulationReaderSettingCategory`:** Distinguishes public and private schools and includes `CensusTract` and an `Unspecified` variant (the `0` or "empty" case).
- **`FIPSParserError`:** Errors specific to parser utilities.

Uses types from Layer 1, concerned with synthetic population specific data parsing and representation.

### Layer 3: Synthetic Population Data Reading

Previously in crate `ixa-aspr`  gated by a feature flag but now incubated in `pop_reader::archive`, we have utilities for reading records from the CSV files in a particular format.

- **`PersonRecord`:** a struct with an age (required) and an `Option<FIPSCode>` field for each of home, school, and work IDs (not required).
- **`PersonRecordIterator`:** A stream of `PersonRecord` values parsed on demand from a given source. Convenience constructors for reading a particular file path.
- **`PopulationReaderError`:** High-level error type, wraps `FIPSParserError` as a source for errors parsing individual fields, with other variants for IO errors, `ZipError` wrapper, and `PopulationReaderError::WrongFieldCount` for malformed rows. Row-local errors are recoverable—parser moves to the next line. All others are trap states, and `next` will return `None`.
- Transparently handles reading files within a zip archive or plain files on the filesystem.
- Very efficient, byte-oriented non-copying parser specifically for the format used in this dataset.
- Useful for any synthetic population dataset in the same format.
- Specific conveniences for the ASPR dataset in particular have been removed.

## Issues and Questions

- Convenience items specific to the ASPR synthetic population dataset have been removed. This code is generally compatible with any dataset in the same format.

- Making the synthetic population reader code actually parse setting IDs into `FIPSCode` values also makes it more brittle, as it is parsing *more* structure. This PR makes the `create_synthetic_population.py` script emit more structured data to satisfy the new parser. So we have increased code complexity in two different places in the code. 

  The (short term) motivation for doing this is, `SettingCode` (now ~an alias~ a newtype for `FIPSCode`) now carries its `SettingCategory` with it, and so the setting category can be determined by inspection of the value rather than us needing to store this information elsewhere.  (We don't take advantage of this in this PR—that's deferred to another PR.)

  But we are parsing a lot more structure than just setting category, namely the state, county, and tract data, and we aren't using it, and we don't have any near term plans to use it, as far as I know. This is not necessarily a problem, but it's something to at least be aware of.

- Related to the previous: There's no getting around the fact that *parsing more structure* takes more time. Still, my optimized parser is fast: init runs ~10% *faster* than in the original implementation.

- The `FIPSCode` type can represent things that carry a FIPS hierarchical geographic region code down to the `tract` granularity with three additional fields: setting category (4 bits), sequential ID (15 bits), and "data" (8 bits, currently unused). This data type might be overspecified. I argue that it's fine for right now, and if we need something different later, we make something different later.

- I have tests that assume the existence of 

  - a zip file containing the zipped ASPR synth pop dataset
  - the unzipped ASPR synth pop dataset

  These tests are gated behind the `aspr_zip_tests` and `aspr_dataset_tests` feature flags respectively, which features are disabled by default. I run them on my local dev machine.

  We should probably set the default "data directory" as defined in `src/pop_reader/archive.rs:DEFAULT_DATA_PATH` to the value of an environment variable at compile time to point to the dataset on the developer's machine. Then we can enable these tests by default.

- ~Are there better names for things I've named `ASPR`?~ Renamed.

### DIFF
--- a/src/pop_reader/archive.rs
+++ b/src/pop_reader/archive.rs
@@ -1,6 +1,24 @@
 /*!
 This module provides facilities to read CSV files in the supported person record format.
 
+The CSV format is as follows:
+
+```csv
+age,homeId,schoolId,workplaceId
+33,56000000300000001,,
+37,56000000300000001,,56000000500000017
+8,56000000300000001,56000000400000001,
+5,56000000300000001,56000000400000001,
+```
+
+The first line is a header, and the remaining lines contain the data. The `age` field is required.
+The `homeId` field may be required by the model and is always present in the datasets we use, but
+the `pop_reader::archive` module does not enforce this. The `schoolId` and `workplaceId` fields
+are optional. The fields are separated by commas without spaces and are unquoted. The ID fields
+must be in the format described in the [`pop_reader`] module-level documentation. The record format
+is parsed by [`PersonRecordIterator::next()`], which itself calls out to parselets for each field,
+while `EXPECTED_FIELD_COUNT` defines the number of fields in each record.
+
 Throughout this module, source data files are addressed using two paths:
 
 - the **data path**, configured with [`set_data_path`] and retrieved with [`get_data_path`]
@@ -500,6 +518,10 @@ fn parse_optional_field<T>(
     Ok(Some(value))
 }
 
+/// An iterator-like structure that produces the fields of a line of CSV data.
+///
+/// This type emits an error message if the row has a number of fields different from
+/// `EXPECTED_FIELD_COUNT`.
 struct FieldCursor<'a> {
     rest: &'a [u8],
     line_number: usize,
--- a/src/pop_reader/archive.rs
+++ b/src/pop_reader/archive.rs
@@ -1,9 +1,9 @@
 /*!
-This module provides facilities to read CSV files in the ASPR person record format.
+This module provides facilities to read CSV files in the supported person record format.
 
 Throughout this module, source data files are addressed using two paths:
 
-- the **data path**, configured with [`set_aspr_data_path`] and retrieved with [`get_aspr_data_path`]
+- the **data path**, configured with [`set_data_path`] and retrieved with [`get_data_path`]
 - the **file path**, which is always interpreted relative to that data path
 
 The data path can have one of two forms:
@@ -16,40 +16,32 @@ The same file path is interpreted differently depending on which kind of data pa
 - if the data path is a directory, the final source file is `data_path.join(file_path)`
 - if the data path is a zip archive, the file path is the name of the file inside the archive
 
-For example, when reading the ASPR synthetic population dataset itself, if the file path is
-`all_states/ak.csv`, then:
+For example, when reading a dataset, if the file path is `all_states/ak.csv`, then:
 
 - with data path `/path/to/ASPR_Synthetic_Population`, the source file is
   `/path/to/ASPR_Synthetic_Population/all_states/ak.csv`
 - with data path `/path/to/ASPR_Synthetic_Population.zip`, the source file is the archive member
   named `all_states/ak.csv`
 
-Set and get the ASPR data path with the `set_aspr_data_path` and `get_aspr_data_path` functions:
+Set and get the data path with the `set_data_path` and `get_data_path` functions:
 
 ```rust,ignore
-let current_path = get_aspr_data_path();
-println!("The current ASPR data path: {:?}", current_path);
+let current_path = get_data_path();
+println!("The current data path: {:?}", current_path);
 
-set_aspr_data_path(PathBuf::from("../CDC/data/ASPR_Synthetic_Population.zip"));
-let new_path = get_aspr_data_path();
-println!("The new ASPR data path: {:?}", new_path);
+set_data_path(PathBuf::from("../CDC/data/ASPR_Synthetic_Population.zip"));
+let new_path = get_data_path();
+println!("The new data path: {:?}", new_path);
 ```
 
-For ASPR-specific conveniences such as [`ALL_STATES_DIR`], [`CBSA_ALL_DIR`],
-[`CBSA_ONLY_RESIDENTS_DIR`], [`NON_CBSA_RESIDENTS_DIR`], and [`MULTI_STATE_DIR`], the assumption is
-that the data path is either:
-
-- `path/to/root/of/unzipped/aspr/dataset`
-- `path/to/ASPR_Synthetic_Population.zip`
-
 You can iterate over the records in a CSV file by passing a file path relative to the configured
-data path to [`ASPRRecordIterator::from_path`]:
+data path to [`PersonRecordIterator::from_path`]:
 
 ```ignore
-# use ixa_aspr::archive::{CBSA_ALL_DIR, ASPRRecordIterator};
+# use crate::pop_reader::archive::PersonRecordIterator;
 # use std::path::PathBuf;
-let file_path = PathBuf::from(CBSA_ALL_DIR).join("AK/Ketchikan AK.csv");
-let records = ASPRRecordIterator::from_path(file_path);
+let file_path = PathBuf::from("cbsa_all_work_school_household").join("AK/Ketchikan AK.csv");
+let records = PersonRecordIterator::from_path(file_path);
 // Do something with the records...
 ```
 
@@ -57,25 +49,16 @@ If the data path is the root of an unzipped dataset, that file path refers to
 `cbsa_all_work_school_household/AK/Ketchikan AK.csv` under that directory. If the data path is a zip
 archive, the same file path refers to the archive member with that name.
 
-The [`ASPRRecordIterator::state_population`] convenience method reads the file whose file path is
-`all_states/{state}.csv` relative to the configured data path. This is an ASPR-specific convenience
-and therefore assumes the data path points at either the root of the unzipped ASPR synthetic
-population dataset or the corresponding zip archive.
-
-```ignore
-# use ixa_aspr::archive::{ASPRRecordIterator};
-# use ixa_fips::USState;
-let records = ASPRRecordIterator::state_population(USState::AK);
-// Do something with the records...
-```
-
 You can get a list of CSV file paths in a given subdirectory with [`iter_csv_files`]. The returned
 paths are always relative to the configured data path, so they can be passed directly to
-[`ASPRRecordIterator::from_path`] or [`ASPRRecordIterator::from_file_iterator`]:
+[`PersonRecordIterator::from_path`] or [`PersonRecordIterator::from_file_iterator`]:
 
 ```ignore
-# use ixa_aspr::archive::{iter_csv_files, ALL_STATES_DIR, ASPRRecordIterator};
-let records = ASPRRecordIterator::from_file_iterator(iter_csv_files(ALL_STATES_DIR).unwrap());
+# use crate::pop_reader::archive::PersonRecordIterator;
+# use crate::pop_reader::archive::iter_csv_files;
+let records = PersonRecordIterator::from_file_iterator(
+    iter_csv_files("all_states").unwrap(),
+);
 // Do something with the records...
 ```
 */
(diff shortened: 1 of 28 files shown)


RobertJacobsonCDC:
# To Do Before Merging

- [X] Rebase.
- [ ] Specific confirmation/feedback on items under "Issues and Questions"
  - [X] ASPR-specific conveniences OK?
  - [ ] Test strategy for tests needing data files
  - [x] Better names for ASPR-named things
- [x] Remove references to ASPR

src/population_loader.rs
```diff
@@ -27,90 +21,90 @@ impl_property!(Age, Person);
...
+        home_id.county_code(),
+        home_id.census_tract_code(),
+        ASPRSettingCategory::CensusTract.into(),
+    ).unwrap()
+}
 
 fn create_person_from_record(
     context: &mut Context,
-    person_record: &PeopleRecord,
+    person_record: ASPRPersonRecord,
```
confunguido: I don't think we should call this record ASPRPersonRecord. Why not just PersonRecord? In general, I think what's particular about ASPRs synthetic population isn't the format but the way they generated the synth population, so I wouldn't fixate on that name for our model.

src/pop_reader/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+pub type IdCode = u16;
+/// The numeric type used for the data code fragment; `u16`
+pub type DataCode = u16;
+
+
+/// A record representing a person in the ASPR synthetic population dataset.
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Default, Debug)]
+pub struct ASPRPersonRecord {
+    pub age: u8,
+    pub home_id: Option<FIPSCode>,
```
confunguido, line 66: Home shouldn't be an option but an error if not in the record. Every person in the synth population should have a home.

RobertJacobsonCDC:
# Annoying consequences of replacing `SettingCode` with `FIPSCode`

This PR replaces the original definition of `SettingCode`, moving from a newtype for `usize` to a newtype for `FIPSCode`. (The first version used a type alias, but that uncovered [a nasty bug in ixa](https://github.com/CDCgov/ixa/issues/843). The rebase uses a newtype.) The model code makes very little use of the _values_ of `SettingCode`, so in terms of consequences to the model code itself this is a pretty noninvasive change. But there ARE downstream consequences.

1. A really annoying consequence of this is that we can no longer use dummy values like `SettingCode(1)`, `SettingCode(2)`, etc., in unit tests, because `FIPSCode` values have much more structure. Right now the PR just defines some simple helpers in the `test` submodules that parse a string representation to a `SettingCode`, so you can use strings taken straight from a synth pop data file. Once you have a few, you end up just copying and pasting for new unit tests... but this seems messy. A cleaner approach would be to have some helper functions in the `pop_reader` module enabled for test builds that can generate random `SettingCode` values. Of course, any code that needs specific state, county, and/or tract info can just use the standard constructor methods for `FIPSCode`, so we're only really talking about when we want something but don't care care what. Maybe I can do this in a later PR, but I figured it can be out of scope for now.
2. The fact that it is a newtype is a little annoying. What we can do in a later PR is implement convenience methods, like a getter for `SettingCategory` (because `FIPSCode` actually encodes `PopulationReaderSettingCategory`, which distinguishes public and private schools and includes a "zero"/"unspecified" value).

src/pop_reader/archive.rs
```diff
@@ -0,0 +1,864 @@
...
+/// chosen to be large enough to keep rebuffering to a minimum, which in turn minimizes the
+/// probability that a line spans a buffer window boundary and thus has to be copied to scratch
+/// memory for processing.
+const ASPR_READER_CAPACITY: usize = 256 * 1024;
+/// The number of fields in each ASPR record: age, homeId, schoolId, workplaceId.
+const ASPR_EXPECTED_FIELD_COUNT: usize = 4;
+
+/// ASPR synthetic population dataset directory name, relative to the dataset root or zip archive
+/// root, containing one CSV file per state.
+pub const ALL_STATES_DIR: &str = "all_states";
```
confunguido: My guess is that this directory structure is optional and that we don't have to define our input files in a different way that we are now. Am I correct?

confunguido (commented):
I made a couple of comments from a partial review. I think I would try to make this less specific to ASPR. It doesn't need a prefix in my opinion.

src/pop_reader/mod.rs
RobertJacobsonCDC: I do raise the error, but I raise it at _person add_ time rather than at record read time. My reasoning is, this lets the reader code work with datasets that allow missing home ID if there ever is such a thing, and the model code can decide if that's an error. It's really just a question of what part of the code should raise the error and how much validation of the input data should happen in the reader code itself. But if it makes the most sense to enforce the existence of home ID in the reader, we can do that. After all, we are requiring that age exist in the reader.

src/pop_reader/archive.rs
RobertJacobsonCDC: Yes. This is an ASPR-specific artifact. I'll remove it with the rest of the ASPR-specific convenience items like we discussed.

src/settings.rs
erik-rosenstrom: Error from that has slipped through the cracks but this should be a ModelError

src/settings.rs
```diff
@@ -50,7 +47,9 @@ impl<T> IndexMut<SettingCategory> for [T; SETTING_COUNT] {
 pub const SETTING_COUNT: usize = SettingCategory::COUNT;
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Copy, Hash)]
-pub struct SettingCode(pub usize);
+pub struct SettingCode(pub FIPSCode);
+// See #843:
```
erik-rosenstrom: Nit: These comments seem like they can be removed

confunguido (approved):
I think this is fine. It is possible that we may want to revisit the structure if we attempt to use a different synth pop dataset with a different format.

src/pop_reader/parser.rs
```diff
@@ -0,0 +1,831 @@
+#![allow(mismatched_lifetime_syntaxes)]
```
erik-rosenstrom, line 1: It could be helpful to have some documentation about what would need to change if we added a new column to a synthetic population file

src/pop_reader/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// A `PopulationReaderSettingCategory` is not itself a FIPS code, but it is implicit in this population record format.
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Default, Debug, FromRepr)]
+#[repr(u8)]
+pub enum PopulationReaderSettingCategory {
```
erik-rosenstrom, line 92: Setting categories have been important for us. So, when we eventually use `PopulationReaderSettingCategory` to replace it we need to be careful to make sure that the code is organized such that is easily connectable to the settings module. Right now, it seems disparate. But maybe that just because all of this is so new

src/settings.rs
erik-rosenstrom: I see this is related to the bug you identified in the PR description. Is it common practice to leave a comment indicating implementation someone might try would lead to a non-obvious bug?

erik-rosenstrom (approved):
I think it looks good and is nice to use from ixa-epi-covid. My main comment is about documentation.

src/settings.rs
RobertJacobsonCDC: Sometimes it's a good idea to leave a comment explaining why something was done a certain way and referencing an issue. In this case, though, it was mostly just for myself and is probably "litter" that should be removed.

src/pop_reader/mod.rs
RobertJacobsonCDC: Yes, well spotted! In a follow-up PR I connect the reader-specific `PopulationReaderSettingCategory` to the model-specific `SettingCategory` in a much nicer way by implementing the method `SettingCode::category()` that returns a `SettingCategory`. In the implementation of that method you would wire the `PopulationReaderSettingCategory` variants to the `SettingCategory` variants.

src/pop_reader/parser.rs
RobertJacobsonCDC: I noticed we're also missing a description of the expected row format. I'll add it to the module-level docs in `src/pop_reader/archive.rs`. (The formats of the ID fields are documented in `src/pop_reader/mod.rs`.)

PR #89, opened 2026-04-09
feat: New SettingMembership index, remove SettingCategory property

RobertJacobsonCDC:
This PR is stacked on top of #84. The major contributions of the PR are:

- Replaces the `Setting` entity with a `SettingMembership` data plugin. 
  - The derived properties for `HomeId`, `SchoolId`, etc., still exist so you can get them with `context.get_property()`. But the data plugin is used for sampling from a setting and for getting the size of a setting. Thus, the derived properties are not indexed.
  - The `SettingMembership` initializes its index upon first access. No need to explicitly call `context.initialize_setting_size()`.
- Many operations no longer require a `SettingCategory` parameter, either because `SettingCode` knows its category, or because the settings are treated uniformly (don't require category-specific treatment).
- Adds convenience methods on `SettingCode` specificaly for constructing dummy values for unit tests (conditionally compiled with `#[cfg(test)]`) and fixes up the unit tests to uses these methods, fixing an annoyance introduced by #84

Minor housekeeping tasks:

- removed or modified obsolete unit tests
- removed `--no-stats` flag in the `run` command of `Makefile`

Local benchmark results:

```
======================================================================
  RobertJacobsonCDC_aspr_reader (2ef36cc) vs RobertJacobsonCDC_setting_code_value
======================================================================

  Benchmark                                      Base         Current  Change
  ---                                             ---             ---  ---
  init/population/10k                       10.704 ms       3.6282 ms  -66.1% faster ✓
  init/population/100k                      134.54 ms       51.878 ms  -61.4% faster ✓
  execute/population/10k                    90.659 ms       83.706 ms  -7.7% faster ✓
  execute/population/100k                    1.3253 s        1.2449 s  -6.1% faster ✓

======================================================================
```

### DIFF
--- a/src/setting_code.rs
+++ b/src/setting_code.rs
@@ -1,0 +1,126 @@
+//! Setting codes are used to identify settings in the population. The `SettingCode` type is
+//! a newtype for `FIPSCode`. Thus, it knows its setting category, state, county, and tract.
+
+use crate::pop_reader::{FIPSCode, PopulationReaderSettingCategory};
+use crate::settings::SettingCategory;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Copy, Hash)]
+pub struct SettingCode(pub FIPSCode);
+
+impl SettingCode {
+    pub fn category(&self) -> SettingCategory {
+        match PopulationReaderSettingCategory::from_repr(self.0.category_code()) {
+            Some(PopulationReaderSettingCategory::Home) => SettingCategory::Home,
+            Some(PopulationReaderSettingCategory::PrivateSchool)
+            | Some(PopulationReaderSettingCategory::PublicSchool) => SettingCategory::School,
+            Some(PopulationReaderSettingCategory::Workplace) => SettingCategory::Work,
+            Some(PopulationReaderSettingCategory::CensusTract) => SettingCategory::Community,
+            _ => panic!("Invalid setting category code: {}", self.0.category_code()),
+        }
+    }
+
+    pub fn extract_community(&self) -> Self {
+        let home_id = self.0;
+        // Since we are calling this constructor with values that we know are valid, we can unwrap.
+        SettingCode(
+            FIPSCode::with_category(
+                home_id.state_code(),
+                home_id.county_code(),
+                home_id.census_tract_code(),
+                PopulationReaderSettingCategory::CensusTract.encode(),
+            )
+            .unwrap(),
+        )
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::pop_reader::PopulationReaderSettingCategory;
+    use crate::pop_reader::fips_code::ExpandedFIPSCode;
+    use crate::pop_reader::states::USState;
+    use std::sync::atomic::AtomicU16;
+
+    // Helper methods to generate arbitrary SettingCodes for unit tests
+    impl SettingCode {
+        fn next_home_code() -> u16 {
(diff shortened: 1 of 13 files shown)


RobertJacobsonCDC:
This PR also cuts memory usage by 40%. First is this PR, second is the `main` branch.

```
━━━━ Execution Summary ━━━━
Max memory usage:        598.2 MiB
Max plans in flight:     1884632
Max plan queue memory:   132.0 MiB
CPU time:                8s 639ms
Wall time:               8s 656ms 406us 334ns

`main`
━━━━ Execution Summary ━━━━
Max memory usage:        1008.9 MiB
Max plans in flight:     1884632
Max plan queue memory:   132.0 MiB
CPU time:                9s 357ms
Wall time:               9s 385ms 343us 125ns
```

src/setting_code.rs
```diff
@@ -0,0 +1,126 @@
+//! Setting codes are used to identify settings in the population. The `SettingCode` type is
+//! a newtype for `FIPSCode`. Thus, it knows its setting category, state, county, and tract.
+
```
erik-rosenstrom, line 3: Is it reasonable to now create a settings folder with both `settings.rs` and `setting_code.rs`

src/settings.rs
```diff
@@ -6,17 +7,79 @@ use std::{
...
-    pop_reader::FIPSCode,
-    population_loader::{CommunityId, HomeId, Person, PersonId, SchoolId, WorkId},
+    population_loader::{ItineraryRatios, Person, PersonId, SettingIds},
+    setting_code::SettingCode,
 };
 
 define_rng!(SettingRng);
 
-define_entity!(Setting);
+// define_entity!(Setting);
```
erik-rosenstrom, line 19: I feel like we will want to leave settings as entities, for now with the property of `SettingCode`. They will likely have other properties, like `isOpen` that will be used in Phase 2 for policies.

erik-rosenstrom:
On my GFE I get the following benchmarking 
```
======================================================================
  main (ab54d88) vs RobertJacobsonCDC_setting_code_value
======================================================================

  Benchmark                                      Base         Current  Change
  ---                                             ---             ---  ---
  init/population/10k                       16.981 ms       6.8789 ms  -59.5% faster ✓
  init/population/100k                      197.66 ms       79.374 ms  -59.8% faster ✓
  execute/population/10k                    154.63 ms       135.53 ms  -12.4% faster ✓
  execute/population/100k                    2.7115 s        2.4950 s  -8.0% faster ✓

======================================================================
```

erik-rosenstrom (changes requested):
This is looking good to me. I do think we should try and keep settings as entities for now.

confunguido (approved):
lgtm

src/settings.rs
RobertJacobsonCDC: Yes, but right now you don't have any data to attach to that entity. Or are you saying we should store the `SettingId` (`EntityId<Setting>`) in the `SettingIds` property of `Person` and then look up the `SettingCode` (equivalently the `SettingCategory`) when we need it?

I think my argument in this PR is, there are overhead costs to using an entity for settings (60% runtime for initialization, ~10% runtime for entire simulation, 170% memory use), but in exchange for paying those costs we get:
- entity creation and property change events
- the query system and indexing system

I say let's not pay these costs if we aren't using these features. We have an API that hides the backing implementation. When we are ready for a more sophisticated implementation, we can change that backing implementation without affecting things outside of the `settings` module.

erik-rosenstrom (approved):
Looks good to me. Could you take a look at creating a settings folder before merging?

PR #106, opened 2026-04-28
chore: Parameter uncertainty md

### LINKED ISSUE #49: Make parameter uncertainty table
Take the list of current parameters in the model and make a table that supplies either a prior or value for those parameters. It would be preferable to specify a distribtuion for every "value" to account for uncertainty and possibly supply literature estimates as priors.

KOVALW:
A single markdown file associated with the phase 1 calibration experiment to track decision making across milestones


eqmooring (changes requested):
Aside from changes proposed in #121 here are a few things not related to symptom status / clinical progression:

- It might be good to more explicitly mirror the contents of `default_params.json`. For example, `max_time` and `initial_prevalence` aren't included here, which could confuse some users.
- Ideally, there would be a reference for the generation interval maximum (duration).
- I don't think a detailed discussion is necessary, but it might be worth a bit more of a comment about the limitations of relating $R_0$ to the rate function (given how density dependence and population structure plays into this). Just highlight that this is non-trivial / makes various assumptions. Also, any reference for 2.1? It's a plausible value, so I'm not too concerned.
- I wonder if maybe some prefatory remarks are needed in this document to explain the goal of this document? I think some readers may see it as redundant to the documents in `docs` / wonder why it lives here versus there.

experiments/phase1/reports/parameter_description.md
```diff
@@ -0,0 +1,71 @@
...
+| `prevalence_report`           | Configuration for the prevalence report, including whether to write, filename, and period.     |
+| `incidence_report`            | Configuration for the incidence report, including whether to write, filename, and period.      |
+| `transmission_report`         | Configuration for the transmission report, including whether to write and filename.            |
+| `aggregated_deaths_report`    | Configuration for the aggregated deaths report, including whether to write, filename, and period.|
+| `first_death_terminates_run`  | Boolean indicating whether the simulation terminates upon the first observed death.            |
+
+Because we specify a constant rate function for infectiousness, an approximation to $R_0$ can be calculated by multiplying the rate by the duration. This value yields the minimum total number of infection attempts that each infectious individual will make from their exposure time until they recover. This is a lower bound on $R_0$ because density-dependent transmission specified by settings properties can lead to higher infeciton attempt emission events as a function of the scaling factor and number of contacts in a setting.
```
eqmooring:
```suggestion
Because we specify a constant rate function for infectiousness, an approximation to $R_0$ can be calculated by multiplying the rate by the duration. This value yields the minimum total number of infection attempts that each infectious individual will make from their exposure time until they recover. This is a lower bound on $R_0$ because density-dependent transmission specified by settings properties can lead to higher infection attempt emission events as a function of the scaling factor and number of contacts in a setting.
```
Is it strictly a lower bound? Aside from density dependence, are there other ways that population structure can become an issue here? I am imagining a synthetic population in which each person is in a 2-person household and the overwhelming majority of time is spent in the household (and a small amount in the community). If my intuition is right, R_0 would be < rate * duration.

experiments/phase1/reports/parameter_description.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Parameter Values
+
+*See `symptom_status_params.qmd` for discussion of how delay distribution parameters were derived.*
+
+| Parameter                                     | Value or Prior                       | Value Source |
+|-----------------------------------------------|--------------------------------------|--------------|
+| `infectiousness_rate_fn (rate)`               | `LogNormal (mean -1.5, std_dev 0.5)` | Scaling by duration, $R_0$ has a mean value of 3.0 and a mode of 2.1 in settings without density-dependent transmission based on the wild-type variant and early reviews like [Billah et al 2020](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0242128&fbclid=IwAR1E_q7tqgAG9DMTA8J6WTlmpEWt2JCje15qyqk1mBdW3vbl1zSBypyRVjY) |
```
eqmooring:
```suggestion
| `infectiousness_rate_fn (rate)`               | `LogNormal (mean -1.5, std_dev 0.5)` | Scaling by duration, $R_0$ has a mean value of 3.0 and a mode of 2.1 in settings without density-dependent transmission based on the wild-type variant and early reviews like [Billah et al 2020](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0242128) |
```
Minor pedantic change to clean up URL.

experiments/phase1/reports/parameter_description.md
```diff
@@ -0,0 +1,71 @@
...
+| `critical_to_resolved_delay (mu)`             | `2.565`                              | [CDC COVID-19 Pandemic Planning Scenarios](https://archive.cdc.gov/www_cdc_gov/coronavirus/2019-ncov/hcp/planning-scenarios.html) |
+| `critical_to_resolved_delay (sigma)`          | `0.814`                              | [CDC COVID-19 Pandemic Planning Scenarios](https://archive.cdc.gov/www_cdc_gov/coronavirus/2019-ncov/hcp/planning-scenarios.html) |
+| `settings_properties (Home)`                  | `Uniform(0, 1)`                      | No prior belief on at-home density-dependent transmission (i.e., all values from extremes of complete density dependence to complete frequency dependence equally likely) |
+| `settings_properties (Work)`                  | `{"alpha": 0.0}`                     | No density-dependent transmission outside the home, following [Hladish et al. 2024](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012128#pcbi.1012128.s001) |
```
eqmooring:
```suggestion
| `settings_properties (Work)`                  | `{"alpha": 0.0}`                     | No density-dependent transmission outside the home, following [Hladish et al. 2024](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012128) |
```

experiments/phase1/reports/parameter_description.md
```diff
@@ -0,0 +1,71 @@
...
+| `settings_properties (School)`                | `{"alpha": 0.0}`                     | No density-dependent transmission outside the home, following [Hladish et al. 2024](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012128#pcbi.1012128.s001) |
```
eqmooring:
```suggestion
| `settings_properties (School)`                | `{"alpha": 0.0}`                     | No density-dependent transmission outside the home, following [Hladish et al. 2024](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012128) |
```

experiments/phase1/reports/parameter_description.md
```diff
@@ -0,0 +1,71 @@
...
+| `settings_properties (Community)`             | `{"alpha": 0.0}`                     | No density-dependent transmission outside the home, following [Hladish et al. 2024](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012128#pcbi.1012128.s001) |
```
eqmooring:
```suggestion
| `settings_properties (Community)`             | `{"alpha": 0.0}`                     | No density-dependent transmission outside the home, following [Hladish et al. 2024](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1012128) |
```

eqmooring (approved):
Looks good to me. I have one comment / question on R_0, and some typographic / stylistic suggestions.

experiments/phase1/reports/parameter_description.md
KOVALW: You’re right that this is not a lower bound on $R_0$. This is really a lower bound on the maximum infection attempt rate, which as you point out is realized as $R_0$ only once we account for the discrete number of contacts on a persons contact network. I’ll change this language

PR #113, opened 2026-05-04
feat: Itinerary manager to enable itinerary modification

erik-rosenstrom:
This PR implements an itinerary manager which modifies an individual's itinerary based on their person properties. The itinerary manager defines an itinerary modifier struct which has elements ranking, itinerary ratios, and modifier type. These modifiers are stored in a data plugin. The data plugin consists of a hashmap with the structure `HashMap<Property<Person>, HashMap<Property<Person>::CanonicalValue, Vec[ItineraryModifier]>>` This data structure enables multiple itinerary modifiers to be registered with a single person property. The following methods are implemented in a context trait extension to interact with this data plugin

- `register_itinerary_modifier`: given a person property and value like `Age(11)` and an itinerary modifier, this method stores the person property and value and itinerary modifier in the data plugin. If an entry exists for the property and value the itinerary modifier is appended to the end of the vector.
- `remove_itinerary_modifier_by_property`: removes the entire vector of itinerary modifiers associated with a property and value. This does not remove the property entry
- `remove_itinerary_modifier_by_property_and_type`: removes a specific element with `modifier_type` from the vector associated with property and value 
- `get_itinerary_modifiers`: iterates over all property and values in the data plugin and assess if the person of interest matches the property values. It returns a vector of all itinerary modifiers that match the person.
- `get_dominant_itinerary_modifier`: calls `get_itinerary_modifiers` sorts them by ranking and pops the highest ranking modifier

### DIFF
--- a/src/itinerary_manager.rs
+++ b/src/itinerary_manager.rs
@@ -1,0 +1,211 @@
+use ixa::{
+    prelude::*,
+};
+use serde::Serialize;
+use std::{any::{TypeId}, collections::HashMap};
+
+use crate::{Age, settings::ItineraryRatios};
+use crate::population_loader::{Person, PersonId};
+
+define_rng!(DummyRng);
+
+#[derive(Debug, PartialEq, Clone, Serialize, Copy)]
+pub struct ItineraryModifier{
+    ranking: usize,
+    itinerary_ratios: ItineraryRatios
+}
+
+// pub trait ItineraryModifiers: std::fmt::Debug + Any {
+//     /// Return the itinerary of a person based on their properties and the current context.
+//     #[allow(dead_code)]
+//     fn get_itinerary(
+//         &self,
+//         context: &Context,
+//         person_id: PersonId,
+//     ) -> Option<ItineraryModifier>;
+
+//     /// For debugging purposes. The name of the itinerary modifier. The default implementation
+//     /// returns the `Debug` representation of the itinerary modifier struct on which this trait
+//     /// is implemented.
+//     fn get_name(&self) -> String {
+//         format!("{self:?}")
+//     }
+// }
+
+// A newtype wrapper for the itinerary modifiers specified via a hashmap of person
+// property values and itinerary -- i.e., `modifier_key: &[(P::Value, Vec<ItineraryEntry>)]`
+#[allow(dead_code)]
+#[derive(Debug)]
+struct PersonPropertyModifier<'a, P>
+where
+    P: Property<Person> + std::fmt::Debug,
+    P::CanonicalValue: std::hash::Hash + Eq + std::fmt::Debug,
+{
+    property: P,
+    modifiers: HashMap<<P as Property<Person>>::CanonicalValue, ItineraryModifier>,
+    _phantom: std::marker::PhantomData<&'a ()>,
+}
+#[allow(dead_code)]
+impl<'a, P> PersonPropertyModifier<'a, P>
+where
+    P: Property<Person> + std::fmt::Debug,
+    P::CanonicalValue: std::hash::Hash + Eq,
+{
+    fn get(&self, key: &P::CanonicalValue) -> Option<&ItineraryModifier> {
+        self.modifiers.get(key)
+    }
+}
+
+// #[allow(dead_code)]
+// impl<P> ItineraryModifiers for PersonPropertyModifier<'_, P>
+// where
+//     P: Property<Person, CanonicalValue = P> + std::fmt::Debug + 'static,
+//     P::CanonicalValue: std::hash::Hash + Eq,
+// {
+//     fn get_itinerary(
+//         &self,
+//         context: &Context,
+//         person_id: PersonId,
+//     ) -> Option<ItineraryModifier> {
+//         let (_person_property, modifier_map) = self;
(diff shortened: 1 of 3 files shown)


src/itinerary_manager.rs
```diff
@@ -0,0 +1,663 @@
...
+pub enum ItineraryModifierType {
+    SchoolClosure,
+    WorkClosure,
+    Weekend,
+    ShelterInPlace,
+    WeekendAndShelterInPlace,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Copy)]
+pub struct ItineraryModifier {
```
confunguido, line 12: I think at this point these aren't itinerary modifiers but itinerary replacements

src/itinerary_manager.rs
```diff
@@ -0,0 +1,663 @@
...
+use std::{
+    any::{Any, TypeId},
+    collections::HashMap,
+};
+
+use crate::population_loader::{Person, PersonId};
+use crate::settings::ItineraryRatios;
+
+#[derive(Debug, PartialEq, Clone, Serialize, Copy, Eq, Hash)]
+pub enum ItineraryModifierType {
```
confunguido: Why do we need this?

src/itinerary_manager.rs
```diff
@@ -0,0 +1,663 @@
...
+        context.register_itinerary_modifier(Age(11), work_closure_modifier);
+        let p1 = context.add_entity::<Person, _>((Age(11),)).unwrap();
+        assert_eq!(
+            context.get_dominant_itinerary_modifier(p1),
+            Some(work_closure_modifier)
+        );
+    }
+
+    #[test]
+    fn example_with_schools_and_weekends() {
```
confunguido: I don't think your dominance itinerary can actually test "schools and weekends" here. Let's remove this test for now and replace with next iteration of layered modifiers

src/itinerary_manager.rs
```diff
@@ -0,0 +1,663 @@
...
+
+        context.add_plan(8.0, move |context| {
+            context.remove_itinerary_modifier_by_property::<Age>(Age(11));
+            assert_eq!(context.get_dominant_itinerary_modifier(p1), None);
+        });
+        context.execute();
+    }
+
+    #[test]
+    fn example_with_shelter_in_place_and_weekend() {
```
confunguido: Same comment as school.

src/itinerary_manager.rs
```diff
@@ -0,0 +1,663 @@
...
+        context.add_plan(8.0, move |context| {
+            context.remove_itinerary_modifier_by_property::<Age>(Age(21));
+            assert_eq!(context.get_dominant_itinerary_modifier(p1), None);
+            assert_eq!(context.get_dominant_itinerary_modifier(p2), None);
+        });
+        context.execute();
+    }
+
+    #[test]
+    fn test_telework_school_closure_sip_and_weekend_modifiers() {
```
confunguido: I don't think this is testing anything else that registering and removing itineraries, which you already did above. To test the combination of itineraries you need to change the dominant itinerary function that we have right now.

confunguido (changes requested):
I think this is fine. But remove unnecessary tests and I want to understand why we need ItineraryModifierType. Also, pre-commit is failing. Please fix.

src/itinerary_manager.rs
```diff
@@ -0,0 +1,663 @@
...
+            let itinerary_modifier_vec = modifier.get_itineraries(self, person_id);
+            if let Some(itinerary_modifier_vec) = itinerary_modifier_vec {
+                modifiers.extend(itinerary_modifier_vec);
+            }
+        }
+        modifiers
+    }
+
+    // This needs to be here to have access to the concrete context type for the get_itinerary trait method
+    fn get_dominant_itinerary_modifier(&self, person_id: PersonId) -> Option<ItineraryModifier> {
```
confunguido: Give that this is the function that will be used in settings, I think the name should be more general, right now it's tied to the "dominant_itinerary" approach.

src/itinerary_manager.rs
erik-rosenstrom: This is a way select itinerary modifiers from within the vector of itinerary modifiers that might be attached to a person property and property value. I am modifying the code such that when you register another a second itinerary modifier of the same type to a person property and property value it will replace the one that is present.

src/itinerary_manager.rs
erik-rosenstrom: Yes these would be the resulting itinerary under the intervention that changes a persons itinerary

src/itinerary_manager.rs
erik-rosenstrom: I changed it to `get_modified_itinerary`

src/itinerary_manager.rs
confunguido: It sounds like we aren't using that feature at this moment so I would say for now we can remove it and add it with the rest when connecting to person property (if we do it).

confunguido (changes requested):
I don't think we need itinerarymodifiertype at this point. Please remove and then it should be ready to merge.

confunguido (approved):
Looks good as the first skeleton for itinerary managers.

PR #115, opened 2026-05-05
Connecting itinerary manager to settings

erik-rosenstrom:
This PR connects the registered itinerary modifiers to settings and the infection propagation loop. A context extension method is implemented for the itinerary manager that returns the itinerary ratios of a person's dominate itinerary modifier. If there are no itinerary modifiers registered the person default itinerary ratios are returned. This implementation does not modify persons itinerary ratios person properties.

### DIFF
--- a/src/itinerary_manager.rs
+++ b/src/itinerary_manager.rs
@@ -1,0 +1,211 @@
+use ixa::{
+    prelude::*,
+};
+use serde::Serialize;
+use std::{any::{TypeId}, collections::HashMap};
+
+use crate::{Age, settings::ItineraryRatios};
+use crate::population_loader::{Person, PersonId};
+
+define_rng!(DummyRng);
+
+#[derive(Debug, PartialEq, Clone, Serialize, Copy)]
+pub struct ItineraryModifier{
+    ranking: usize,
+    itinerary_ratios: ItineraryRatios
+}
+
+// pub trait ItineraryModifiers: std::fmt::Debug + Any {
+//     /// Return the itinerary of a person based on their properties and the current context.
+//     #[allow(dead_code)]
+//     fn get_itinerary(
+//         &self,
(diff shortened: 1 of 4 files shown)


src/itinerary_manager.rs
```diff
@@ -206,6 +207,51 @@ impl ContextItineraryModifierExt for Context {
...
+
+    fn get_dominant_itinerary_modifier_ratios(&self, person_id: PersonId) -> ItineraryRatios {
+        self.get_dominant_itinerary_modifier(person_id)
+            .map(|modifier| modifier.itinerary_ratios)
+            .unwrap_or(self.get_property::<Person, ItineraryRatios>(person_id))
+    }
+
+}
+
+pub fn init(context: &mut Context) -> Result<(), ModelError> {
```
confunguido, line 219: These itineraries look like a test. It's okay to leave them as part of the tests if we aren't ready to define some itinerary modifiers them right now.

src/itinerary_manager.rs
```diff
@@ -206,6 +207,51 @@ impl ContextItineraryModifierExt for Context {
     fn get_dominant_itinerary_modifier(&self, person_id: PersonId) -> Option<ItineraryModifier> {
         self.get_itinerary_modifiers(person_id).into_iter().max()
     }
+
+    fn get_dominant_itinerary_modifier_ratios(&self, person_id: PersonId) -> ItineraryRatios {
```
confunguido, line 211: This looks like it will change when you incorporate a different approach for the itinerary modifiers.

confunguido (changes requested):
A few comments but also think this needs to wait until we modify the dominant itinerary approach.

erik-rosenstrom:
Closing for now as approach to itinerary modifiers is changing. I will open another branch when we are ready to implement this step.

PR #117, opened 2026-05-06
fix: asymptomatic calculation to be 1 minus symptomatic in importation

KOVALW:
Currently the probability of mild given infect is set to be the asymptomatic probability for imported infections, but it should be 1.0 minus that probability


erik-rosenstrom (approved):
Looks good to me.

PR #123, opened 2026-05-12
feat: Itinerary modifiers as an absorbing Markov chain

erik-rosenstrom:
This PR implements itinerary modifiers, their layering and application to a base itinerary as an absorbing Markov chain. The fundamentals of absorbing Markov chain theory are used. 

This approach requires differentiating interventions that modify a person itinerary as either a change to underlying activities or behaviors and a change to a location of their behavior. This categorization yields an intuitive ordering that changes to activities should be applied before changes to locations of activities. 

We model changes to activities as an absorbing Markov chain. We conceptually model 12 states: `home_activity`, `home_location`, `home_absorb`, `work_activity`, `work_location`, `work_absorb`, `school_activity`, `school_location`, `school_absorb`, `community_activity`, `community_location`, `community_absorb`.  The transition matrix defines the probability of going from the row state to the column state in the next step. Since the transition represent probabilities the sum over the row must equal one. The transition matrix between states can be organized to have four quadrants:

1. Transient to transient states (8x8) referred to as Q or the transient matrix
2. Transient to absorbing states (8x4) referred to as R or the absorbing matrix
3. Absorbing to transient states (4x8), this will always be the 0 matrix
4. Absorbing to absorbing states (4x4), this will always be the identity matrix

The transient states are `_activity` and `_location` states, with the remaining being absorbing. Given the nature of itinerary modifiers and the order of in which we apply changes, we impose a structure on the states/transitions of the Markov chain. This state structure is akin to 3 levels: Activity, Location, Absorbing. Transitions between states of given setting are one minus the sum of all other transitions out of the setting state. Other transitions out of a setting state are defined by an itinerary modifier.  An itinerary modifier only adds transitions between activity or location states.

According to absorbing Markov chain theory, for a base itinerary p (a row vector), the modified itinerary m after applying an activity itinerary modifier is `m = p * (I-Q)^-1 * R`. In this equation (I-Q)-1R results in the probability of being absorbed to a particular absorbing state given which transient state you begin in. Our default itinerary serves as the probability distribution over our starting states (the activity state version of each setting). Multiplying the two together results in the probability we are absorbed into each absorbing state version of the settings

When layering multiple changes to activities, we add Q matrices together and normalize values if the row sum is greater than 1. We calculate the absorbing matrix and modified itinerary for the Q matrix corresponding to the set of layered modifiers. 

An example of a Markov chain for weekends and virtual school closure itinerary modifiers are shown below. Virtual school closure moves the time spent doing schoolwork at school to doing schoolwork at home. Weekends change time spent doing school/work activities to doing activities in the home or community. We will assume that time is split equally between home and the community.

image


src/itinerary_manager.rs
```diff
@@ -206,107 +407,266 @@ mod test {
...
+        assert_eq!(context.get_itinerary_modifiers(p1), vec![weekend_modifier]);
+        assert_eq!(context.get_itinerary_modifiers(p2), vec![weekend_modifier]);
+
+        context.register_itinerary_modifier(Age(11), sip_modifier);
+        context.register_itinerary_modifier(Age(11), itinerary_modifier);
+        context.set_property(
+            p2,
+            ItineraryRatios {
+                itinerary_ratios: [0.3, 0.0, 0.5, 0.2],
+            },
```
confunguido: Shouldn't there be an assertion here?

confunguido (approved):
I am approving because I think it meets the functionality requirements. However, I wonder if we could implement itinerary modifiers in a more generic way, i.e., as a trait, such that `get_modified_itinerary()` would be agnostic to what goes into the modifier as long as it implements the `ItineraryModifier` trait. For instance, if `ItineraryModifier` implements `apply()` then, `get_modified_itinerary` could look like: 
```rust
let modifiers = self.get_itinerary_modifiers(person_id);
let mut itinerary = self.get_property::<Person, ItineraryRatios>(person_id).itinerary_ratios;
for modifier in modifiers {
   itinerary = modifier.apply(&itinerary);
 }
)
```

While the trait could like: 
```rust 
pub trait ItineraryModifier {
  fn apply(&itinerary) -> itinerary(or matrix);
}
```

I can imagine that could design a similar pattern for layered itineraries as long as the trait implement some function such as  `fn layer(&ItineraryModifier) -> ItineraryModifier`

confunguido (approved):
This is fine for now. Please consider my previous request.

eqmooring:
I haven't taken a really deep dive into the code, but I have tried out the math a bunch and am commenting on that basis.

I have sought to understand by way of example how the absorbing Markov approach of creating separate Q and R matrices compares and contrasts with what I would call the "naive" matrix approach where there's a single matrix (equal to the sum of Q and R, at least for some examples) for each itinerary modifier. When dealing with a single modifier, I think these approaches are equivalent. But I *think* that when there are two modifiers of equal priority, adding up and normalizing Q modifies itineraries in a way that logically makes sense, but just adding up the "naive" overall matrices is not necessarily going to yield logical results. It might be good to have some worked examples in documentation that show why this apparently more complex approach has value.

On the topic of distinguishing "activity" versus "location" modification matrices, one example I have been thinking about is quarantine of persons with a known exposure (one could equivalently use isolation as an example, but people who need to isolate are generally sick, so there's a stronger presumption that their activities will have to change): For workers who can telework, quarantine is a change in location, but for workers who cannot telework, quarantine is a change in activity. It's not clear to me how this example would fit into the "activity" versus "location" schema. (Perhaps because interventions of quarantine and isolation involve moving all time to "home", it doesn't really matter.)

I agree that in the case of applying weekends and virtual school, the "activity" versus "location" distinction makes sense and getting the order right is necessary to get logical results. But I wonder if this is a special and not fully generalizable case. (It may be an illustration of how some itinerary modifiers take precedence over others.) Considering a wider variety of examples might help clarify.

It's good to find generalizable approaches if possible, but I also think there's value in allowing for empirical flexibility. To continue the weekends and virtual school example, it's plausible that virtual school as a policy signals risk to families and therefore leads to behavior change on the weekends too (e.g., less time in "community"), even though that's not directly part of virtual school.

erik-rosenstrom:
I will add more detail to the other approaches section of the documentation to demonstrate the drawbacks of other approaches. 

In regards to quarantine, if there is a person property indicating who can and who cannot telework, we can register different itinerary modifiers, accordingly. But I agree, since both implementations will ultimately move all itinerary ratio to home, the difference may be not that meaningful. What may be more meaningful is capturing who is quarantining to calculate the costs associated with an intervention. 

I agree that it is difficult to know if this is fully generalizable to the set of all possible interventions, but for reasonable definitions of interventions we will use in Phase 2, I have not encountered any issues using this method. I will expand examples in the design document to demonstrate this. 

Changes that are coming soon, will separate a modelers implementation of itinerary modifiers from the itinerary manager, giving modelers the ability to modify itinerary modifier logic easily.

Finally, I see your point about virtual school closures impacting the community activity of the family with students or the entire population. This may be an example of when it is beneficial to decompose an intervention's effects into multiple parts and model a community activity reduction separately but in tandem with virtual school closure. I believe this would get at community reductions in weekends.

PR #125, opened 2026-05-12
hotfix: cache proportion data always and set test caching to true

KOVALW:
We currently have no way to get the census data automatically due to API change requests being incompatible with `census` python package. This hotfix caches the state proporiton data used throughout the repo to ensure tests that fail related to this bug can pass

### DIFF
--- a/packages/importation/src/importation/geographies.py
+++ b/packages/importation/src/importation/geographies.py
@@ -39,7 +39,14 @@ def get_total_state_population_data(
         - The function uses the Census API and requires a valid API key.
         - Cached files are stored in the ".cache" directory with filenames in the format
           "state_population_data_<year>.csv".
+    Raises:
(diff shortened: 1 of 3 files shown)


packages/importation/src/importation/geographies.py
```diff
@@ -42,7 +42,7 @@ def get_total_state_population_data(
     """
     filename = f"state_population_data_{year if year else 'latest'}.csv"
```
erik-rosenstrom, line 50: Should we restrict this to 2020 for now rather than letting year be specified since that is the only cached file?

erik-rosenstrom (changes requested):
Looks good, just a question about the year argument

eqmooring (approved):
Looks good. The fact that the checks are passing on this PR is re-assuring.

erik-rosenstrom (approved):
LGTM

PR #127, opened 2026-05-18
chore: method for fips code for the transmission report

### LINKED ISSUE #126: Improving setting code serialization in transmission report
Currently the representation of the setting code in the transmission report is meaningless. We need to record a meaningful version of the code

erik-rosenstrom:
This PR adds a method to make SettingCodes more meaningful in the transmission report. An alternative approach would be rewriting the display implementation for FIPSCodes, but not sure if we want to overwrite the call to displaying the expanded FIPSCode.


confunguido (approved):
lgtm

PR #131, opened 2026-05-20
feat: Interface between itinerary modifiers and itinerary modifier manager

erik-rosenstrom:
This PR separates the implementation of the itinerary modifier logic from how they are stored in the itinerary modifier manager. A trait `ItineraryModifierTrait` is implemented which requires a `layer()` and `apply()` method which combine itinerary modifiers together and apply them to a default itinerary, respectively. The implementation of these methods on an structure that implements the `ItineraryModifierTrait` is done in a separate code module. What is now stored in the itinerary manager data structures and passed between methods are `Box<dyn ItineraryModifierTrait>>`


erik-rosenstrom:
closing in favor of #135

PR #132, opened 2026-05-20
chore: aligning input json to phase 1 calibration

### LINKED ISSUE #87: default input file
Our current input.json may have values that aren't the default values we use in simulations. For instance, all alphas are 0.0 but those aren't the values we use for the calibration.

erik-rosenstrom:
This PR aligns `input.json` with results of phase 1 calibration and [corresponding documentation](https://github.com/CDCgov/ixa-epi-covid/blob/main/experiments/phase1/reports/parameter_description.md).

### DIFF
--- a/input/input.json
+++ b/input/input.json
@@ -3,14 +3,14 @@
         "seed": 1234,
         "max_time": 100.0,
         "synth_population_file": "input/people_test.csv",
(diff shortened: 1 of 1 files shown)


confunguido (approved):
lgtm

PR #133, opened 2026-05-20
chore: Bump ixa to beta2.4

RobertJacobsonCDC:
This PR updates the codebase for `ixa@2.0.0-beta2.4`.

I also snuck in these:

- An optimization in `calculate_multiplier` that avoids `f64::powf` for the common case of $\alpha\in \{0.0, 1.0\}$.
- Made the `profiling` feature no longer enabled by default.

### DIFF
--- a/src/infectiousness_manager.rs
+++ b/src/infectiousness_manager.rs
@@ -1,6 +1,7 @@
-use ixa::{impl_derived_property, prelude::*};
+use ixa::prelude::*;
 use rand_distr::Exp;
 use serde::{Deserialize, Serialize};
+use std::hash::{Hash, Hasher};
 
 use crate::{
     population_loader::PersonId,
@@ -12,7 +13,7 @@ use crate::population_loader::Person;
 
 use ixa::profiling::{increment_named_count, open_span};
 
-#[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Copy)]
+#[derive(Serialize, Deserialize, Debug, Clone, Copy)]
 pub enum InfectionData {
     Susceptible,
     Infectious {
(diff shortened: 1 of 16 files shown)


src/infectiousness_manager.rs
```diff
@@ -26,21 +27,83 @@ pub enum InfectionData {
     },
 }
 
+impl PartialEq for InfectionData {
```
erik-rosenstrom, line 30: Are these `impl`s going to be here long term? If so, it seems reasonable to define `InfectionData` in another file to ensure that logic of the infection manager is clear.

src/population_loader.rs
```diff
@@ -22,11 +23,34 @@ impl_property!(Age, Person);
...
 
-#[derive(Debug, PartialEq, Clone, Copy, Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 pub struct Itinerary {
     pub setting_ids: [Option<SettingCode>; SETTING_COUNT],
     pub itinerary_ratios: [f64; SETTING_COUNT],
 }
+
+impl PartialEq for Itinerary {
+    fn eq(&self, other: &Self) -> bool {
```
erik-rosenstrom, line 33: Same comment as for `InfectionData`

src/symptom_status_manager.rs
```diff
@@ -42,20 +44,158 @@ pub enum SymptomData {
     },
 }
 
-impl_property!(SymptomData, Person, default_const = SymptomData::NoSymptoms);
+impl PartialEq for SymptomData {
```
erik-rosenstrom, line 47: Same as above

src/settings.rs
```diff
@@ -252,9 +252,22 @@ pub trait ContextSettingExt:
     }
 
     fn calculate_multiplier(&self, setting: SettingCode) -> Result<f64, ModelError> {
-        let size = self.get_setting_size(setting) as f64 - 1.0;
         let alpha = self.get_setting_alpha(setting.category())?;
-        Ok(size.powf(alpha))
+        match alpha {
```
erik-rosenstrom, line 256: I forget if you said you weren't seeing speed ups with this optimization. But if you were using something like `bench-compare` all of the alphas are non-trivial values.

erik-rosenstrom (approved):
Looks good. Just few nit comments about the `impl` and code organization. If necessary we can address the organization in a separate issue.

src/settings.rs
RobertJacobsonCDC: You are right that this optimization is very limited, but it's also really simple and low risk, so I figured I'd add it. A separate PR should address this particular bottleneck in a more principled way, so this is just a temporary stopgap anyway.

src/infectiousness_manager.rs
RobertJacobsonCDC: No, I don't think so. We are working on a couple of different alternatives in Ixa core for dealing with this based on our experience updating existing codebases. Let's consider this a temporary measure for now and be conservative with respect to changes in project structure—unless you have other reasons to break these out.

On the other hand, I personally like having things in their own files rather than large files with many things in them.

PR #135, opened 2026-05-21
feat: interface between itinerary manager and itinerary modifiers

erik-rosenstrom:
This PR separates the implementation of the itinerary modifier logic from how they are stored in the itinerary modifier manager. A trait ItineraryModifierTrait is implemented which requires a layer() and apply() method which combine itinerary modifiers together and apply them to a default itinerary, respectively. The implementation of these methods on an structure that implements the ItineraryModifierTrait is done in a separate code module. What is now stored in the itinerary manager data structures and passed between methods are Box<dyn ItineraryModifierTrait>>


src/itinerary_modifiers.rs
```diff
@@ -0,0 +1,230 @@
...
+    activity_matrix: [[f64; SETTING_COUNT]; SETTING_COUNT],
+    location_matrix: [[f64; SETTING_COUNT]; SETTING_COUNT],
+    absorption_probabilities: Option<[[f64; SETTING_COUNT]; TRANSIENT_STATE_COUNT]>,
+}
+
+impl ItineraryModifierTrait for ItineraryModifier {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+    fn layer(&mut self, other: Box<dyn ItineraryModifierTrait>) -> Box<dyn ItineraryModifierTrait> {
```
confunguido, line 19: Do you need to make sure that the incoming modifier is of type matrix?

src/itinerary_modifiers.rs
```diff
@@ -0,0 +1,230 @@
...
+            for i in 0..SETTING_COUNT {
+                new_itinerary[j] += current_itinerary[i] * absorption_probs[i][j];
+            }
+        }
+        new_itinerary
+    }
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Copy)]
+pub struct ItineraryModifier {
```
confunguido: Should you just impl the trait to the itinerarymodifiermatrix?

confunguido (changes requested):
some minor comments

src/itinerary_modifiers.rs
erik-rosenstrom: The line `if let Some(other_modifier) = other.as_any().downcast_ref::<ItineraryTransitionMatrix>()` checks if the argument is an `ItineraryTransitionMatrix`

src/itinerary_modifiers.rs
erik-rosenstrom: Yes. I have done the trait impl on `ItineraryTransitionMatrix` directly

erik-rosenstrom:
Closing this due to accidental rebase to main. Replacing it with #136

PR #141, opened 2026-05-27
Connecting itinerary manager to settings module

### LINKED ISSUE #137: Connecting itinerary modifiers to setting module
During infection propagation, modified itinerary ratios should be used to inform setting sampling and infectiousness calculations

erik-rosenstrom:
This PR connects the itinerary manager module to the settings module. In the setting module method `get_active_settings_for_person` rather than a person's itinerary ratios being accessed via `get_property` the itinerary manager method `get_modified_itinerary` is called. If the returned ratio of a setting is 0, the setting is not included in the set of active settings.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -7,6 +7,7 @@ use std::{
 };
 use strum::{EnumCount as EnumCountMacro, EnumIter, IntoEnumIterator};
 
+use crate::itinerary_manager::ContextItineraryModifierExt;
 pub(crate) use crate::{
     ContextParametersExt, Params,
     error::ModelError,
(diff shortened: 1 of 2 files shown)


src/settings.rs
```diff
@@ -179,13 +184,15 @@ pub trait ContextSettingExt:
     ) -> Result<Vec<(SettingCode, f64, f64)>, ModelError> {
         let mut active_settings = Vec::new();
         let setting_ids = self.get_property::<Person, SettingIds>(person_id);
-        let itinerary_ratios = self.get_property::<Person, ItineraryRatios>(person_id);
+        let itinerary_ratios = self.get_modified_itinerary(person_id);
```
confunguido: Shouldn't this just be `get_itinerary`?

confunguido (approved):
Please make the requested change.

PR #142, opened 2026-05-28
policy and trigger

### LINKED ISSUE #139: Policy and Trigger design
We need a framework to apply interventions to the population. Policy and triggers should be designed with the following considerations:
- The framework should enable specification of start and end criteria. Potential start and end criteria are given as simulation times from the input, or related to the collective state of a group of people (e.g., the prevalence of disease in a school)
- The framework should enable specification of intervention to be used. Ideally, it should be flexible enough to handle itinerary modification, transmission modifiers, surveillance, and in the future potentially vaccines.
- The framework should enable specification of who the intervention will be applied to. Potential ways to specify this are by setting membership or a query of the population.
- Policies should be specifiable from model input. Policy parameterization may need to be informed by some data source (e.g., geographic heterogeneity or human mobility data)

erik-rosenstrom:
This PR implements a policy module. A policy is comprised of a trigger (start and end conditions), an intervention, and a group of people. When the trigger signally the start occurs, the intervention is activated for the group of people, and when the trigger signaling the end occurs, the intervention is deactivated for the group of people. 

Triggers are implemented as an enum. Each value of the enum has its own named fields. Currently there are three implemented. 

- Time trigger: when the simulation reaches the `start_time` the intervention is activated and when it reaches the `end_time` the intervention is deactivated.
- Hospitalization threshold: when the simulation reaches a `threshold` of hospitalizations the intervention is activated.
- On simulation initialization: this activates an intervention at the beginning of the simulation

The logic to trigger the policy is implemented on the triggers enum. Each type of trigger is responsible for emitting a `PolicyTriggerEvent` which has a Boolean field indicating if the policy is active and field for the trigger itself. In the policy interface the trigger of a policy is checked against the trigger of the `PolicyTriggerEvent` to ensure the right policy is executed if multiple policies are in place. The implementation of triggers as an enum is critical development step and when this enum has matured we can investigate implementing it as a trait. 

In this implementation the intervention is generic, so it is implemented as a trait. It has trait methods activate and deactivate, which enable specific activation and deactivation logic for each intervention through a generic interface. 

In this implementation, groups are described by a person property. This enables storage in data structures that map person properties to interventions and enables querying with the person property.


src/policy.rs
```diff
@@ -0,0 +1,553 @@
...
+            modifier: isolation_modifier,
+            active_property: DummyIsolating(true),
+            inactive_property: DummyIsolating(false),
+        };
+        let policy: Policy<DummyIsolating, DummyIsolating, (Age,)> = Policy {
+            trigger: Trigger::Time(TimeTrigger {
+                start_time: 1.0,
+                end_time: 5.0,
+            }),
+            intervention: Intervention::Isolation(isolation_intervention),
```
KOVALW: It seems strange to me to refer to active/inactive properties that are always boolean in the context of setting a modifier for a group over a specified state-based time frame. This is distinct from an individual level policy that has a *trigger* that is tethered to a changing person property rather than a state-based trigger. So we would connect directly to a property or itinerary modifier here instead of an Intervention object.

Individual-level triggers and state-level triggers would allow us to have a more cohesive framework of how policies listen to changes

erik-rosenstrom:
Closing in favor of NPI specific PRs

PR #143, opened 2026-05-29
Removing long string from test

### LINKED ISSUE #45: Update test to not rely on input string
Currently, the test `test_list_reports` relies on the input JSON format being included here as a long string. Consequently, every time updates are made that affect the input parameters, this string needs to be updated, even if there aren't any substantive changes affecting the reports, which are the subject of the test.

The preferred behavior would be to still test aspects of report creation without having to rely on always updating this string whenever variables are added.

https://github.com/CDCgov/ixa-epi-covid/blob/346fc0a8ba08ad9c8230a0e2e537f7a9298d0e85/src/reports/mod.rs#L119

erik-rosenstrom:
This PR addresses #45. Rather than maintain a long string, the default input parameters are used and converted to a string.

### DIFF
--- a/src/reports/mod.rs
+++ b/src/reports/mod.rs
@@ -10,10 +10,12 @@ pub mod incidence_report;
 pub mod prevalence_report;
 pub mod transmission_report;
 
+// the skip_serializing_if is used to avoid having the period field show up in the json in the tests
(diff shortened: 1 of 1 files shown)


KOVALW (approved):
LGTM

PR #146, opened 2026-05-29
feat: adding teachers to in house synthetic populations

### LINKED ISSUE #144: Add teachers to in house synthetic populations
We need to add teachers to schools in `packages/create_synthetic_population/src/create_synthetic_population/run.py`

erik-rosenstrom:
This PR adds teachers to schools in synthetic populations created by our in-house script. 2.56% of adult workers are randomly assigned to schools. According to the national center of education statistics, there are about 3.84 million teachers. There were about 150 million workers in 2020 according to the Bureau and Labor Statistics.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ synthetic-population:
 
 # Run synthetic population generator tests
 test-syn-pop:
-	uv run pytest scripts/test_create_synthetic_population.py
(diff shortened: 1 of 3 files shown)


confunguido (changes requested):
I would prefer to do this in the ixa code with a parameter determining the ratio of students/teachers, such as `student_teacher_ratio`. You can grab either at random or workers from the same county to be reassigned.

PR #158, opened 2026-07-10
Updating ixa version

erik-rosenstrom:
Updating the ixa version for this branch. It removes canonical values and has one small change for multi-property syntax

### DIFF
--- a/src/itinerary_manager.rs
+++ b/src/itinerary_manager.rs
@@ -42,7 +42,7 @@ type PersonPropertyItineraryModifier<'a, P> = (
 
 impl<P> ItineraryModifierStorageTrait for PersonPropertyItineraryModifier<'static, P>
 where
-    P: Property<Person> + std::fmt::Debug + std::hash::Hash + Eq,
+    P: IndexableProperty<Person>,
 {
     fn get_itinerary_modifiers(
         &self,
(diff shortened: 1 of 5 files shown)


src/itinerary_manager.rs
```diff
@@ -80,15 +79,13 @@ define_data_plugin!(
 pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
     /// Register a generic itinerary modifier.
     fn register_itinerary_modifier<
-        P: Property<Person> + std::fmt::Debug + 'static,
+        P: Property<Person> + std::fmt::Debug + std::hash::Hash + Eq + 'static,
```
RobertJacobsonCDC: You can simplify this line to just

```rust
P: IndexableProperty<Person>,
```

To break it down:
- `Property<Person>` already implies `Debug + 'static`.
- `IndexableProperty<Person>` implies `Property<Person> + Eq + Hash`

src/itinerary_manager.rs
```diff
@@ -120,12 +117,12 @@ pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
         }
     }
 
-    fn remove_itinerary_modifier_by_property<P: Property<Person> + 'static>(
+    fn remove_itinerary_modifier_by_property<
+        P: Property<Person> + std::hash::Hash + Eq + 'static,
```
RobertJacobsonCDC: Same as above, can be simplified to

```rust
P: IndexableProperty<Person>,
```

But I think there is a reasonable argument for making the additional `Eq + Hash` requirements explicit. (The reader doesn't have to look up what `IndexableProperty` means if you make it explicit.) In that case, you could write:

```rust
P: Property<Person> + std::hash::Hash + Eq
```

The only difference is, I dropped the `'static`, because `Property<E>` automatically implies `'static`.

As an aside: The Rust community tends to spell out the fully qualified names of things from the standard library a lot more than I personally do. So you see people write things like these random examples:

```rust
let shared = std::sync::Arc::new(String::from("hello"));
std::thread::sleep(std::time::Duration::from_secs(1));
let file = std::fs::File::open("config.txt")?;
```

I personally prefer to just `use std::sync::Arc;` at the top of the file and refer to `Arc` (without the fully qualified name) in the code itself. This is completely stylistic and personal preference.

RobertJacobsonCDC (approved):
Just a couple of redundant trait constraints, and an editorial comment on style.

PR #166, opened 2026-08-03
Acceptance functions for itinerary modifiers and weekends implementation

### LINKED ISSUE #153: Implement school calendar with itinerary modifiers
We need to implement weekends and a school calendar module. Investigate the use of ixa trigger implementation and explore if a functional filter needs to be added to itinerary modifiers so new person properties do not necessarily need to be created

### LINKED ISSUE #168: Acceptance Functions on Itinerary Modifiers
Rather than having complex registration and remove, itinerary modifiers should have a function that should be evaluated to determine if they should be applied or not. This function should take in context, person id and potentially other arguments to make the evaluation.

erik-rosenstrom:
This PR extends itinerary modifiers to only be applied based on the context and person id. Itinerary modifiers are still stored in a hash map with person property values, but now there is an additional way to filter which ones should be applied. The alleviates potentially complicated processes for registering and deregistering them. 
 
This PR adds a new optional field, `acceptance_function` to the `ItineraryTransitionMatrix`. This field is of type `pub type AcceptanceFunction = Rc<dyn Fn(&Context, PersonId) -> bool>;`. So this function has arguments of `context` and `person_id` and returns a Boolean.  The function is evaluated at the point when itinerary modifier is applied to a base itinerary. If true, the itinerary modifier is applied to the base itinerary. If false, the itinerary modifier is not applied to the base itinerary. 

This new feature is used to implement weekends for students, in the school calendar module.

### DIFF
--- a/input/input.json
+++ b/input/input.json
@@ -54,6 +54,11 @@
                             "School": {"alpha": 0.0},
                             "Community": {"alpha": 0.0}},
         "itinerary_ratios": {"Home": 0.25, "Work": 0.25, "School": 0.25, "Community": 0.25},
+        "weekends": {
+            "delay": 3.0,
+            "prop_school_time_to_home": 0.5,
+            "prop_school_time_to_comm": 0.5
+        },
         "prevalence_report": {
             "write": true,
             "filename": "person_property_count.csv",
--- a/input/input.json
+++ b/input/input.json
@@ -54,11 +54,21 @@
                             "School": {"alpha": 0.0},
                             "Community": {"alpha": 0.0}},
         "itinerary_ratios": {"Home": 0.25, "Work": 0.25, "School": 0.25, "Community": 0.25},
-        "weekends": {
-            "delay": 3.0,
-            "prop_school_time_to_home": 0.5,
-            "prop_school_time_to_comm": 0.5
-        },
+        "school_calendar": [
+            {
+                "modifier": "Weekend",
+                "start_time": 3.0,
+                "prop_school_time_to_home": 0.5,
+                "prop_school_time_to_comm": 0.5
+            },
+            {
+                "modifier": "SummerBreak",
+                "start_time": 151.0,
+                "end_time": 226.0,
+                "prop_school_time_to_home": 0.5,
+                "prop_school_time_to_comm": 0.5
+            }
+        ],
         "prevalence_report": {
             "write": true,
             "filename": "person_property_count.csv",
--- a/src/itinerary_manager.rs
+++ b/src/itinerary_manager.rs
@@ -160,13 +160,19 @@ impl ContextItineraryModifierExt for Context {
         for modifier in modifiers {
             layered_modifier = Some(match layered_modifier {
                 Some(mut existing) => {
-                    if modifier.accept(self, person_id){
+                    if modifier.accept(self, person_id) {
                         existing.layer(modifier)
                     } else {
                         existing
                     }
-                },
-                None => modifier,
+                }
+                None => {
+                    if modifier.accept(self, person_id) {
+                        modifier
+                    } else {
+                        continue;
+                    }
+                }
             });
         }
         if let Some(mut layered_modifier) = layered_modifier {
@@ -179,12 +185,14 @@ impl ContextItineraryModifierExt for Context {
 
 #[cfg(test)]
 mod test {
+    use std::sync::Arc;
+
     use super::*;
     use crate::Age;
-    use crate::itinerary_modifiers::define_itinerary_modifier;
+    use crate::itinerary_modifiers::{AcceptanceFunction, define_itinerary_modifier};
     use crate::parameters::{GlobalParams, Params, SettingProperties};
     use crate::settings::{Itinerary, SettingCategory};
-    use ixa::HashMap;
+    use ixa::{ExecutionPhase, HashMap};
 
     fn setup() -> Context {
         let mut context = Context::new();
@@ -223,7 +231,8 @@ mod test {
             [0.5, 0.0, 0.0, 0.5],
             [0.0, 0.0, 0.0, 0.0],
         ];
-        let weekend_modifier = define_itinerary_modifier(Some(weekend_transient_matrix), None, None);
+        let weekend_modifier =
+            define_itinerary_modifier(Some(weekend_transient_matrix), None, None);
 
         context.register_itinerary_modifier(Age(11), weekend_modifier.clone());
         let p1 = context.add_entity(with!(Person, Age(10))).unwrap();
@@ -258,7 +267,8 @@ mod test {
             [0.0, 0.0, 0.0, 0.0],
         ];
 
-        let weekend_modifier = define_itinerary_modifier(Some(weekend_transient_matrix), None, None);
+        let weekend_modifier =
+            define_itinerary_modifier(Some(weekend_transient_matrix), None, None);
         let school_transient_matrix = [
             [0.0, 0.0, 0.0, 0.0],
             [0.75, 0.0, 0.0, 0.25],
@@ -273,10 +283,12 @@ mod test {
         let modifiers = context.get_itinerary_modifiers(p1);
         assert_eq!(modifiers.len(), 2);
         assert!(
-            modifiers.contains(&(Box::new(school_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
+            modifiers
+                .contains(&(Box::new(school_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
         );
         assert!(
-            modifiers.contains(&(Box::new(weekend_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
+            modifiers
+                .contains(&(Box::new(weekend_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
         );
     }
 
@@ -298,13 +310,15 @@ mod test {
         let modifiers_p1 = context.get_itinerary_modifiers(p1);
         assert_eq!(modifiers_p1.len(), 1);
         assert!(
-            modifiers_p1.contains(&(Box::new(weekend_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
+            modifiers_p1
+                .contains(&(Box::new(weekend_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
         );
 
         let modifiers_p2 = context.get_itinerary_modifiers(p2);
         assert_eq!(modifiers_p2.len(), 1);
         assert!(
-            modifiers_p2.contains(&(Box::new(weekend_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
+            modifiers_p2
+                .contains(&(Box::new(weekend_modifier.clone()) as Box<dyn ItineraryModifierTrait>))
         );
 
         // This would remove all age based itinerary modifiers that is not ideal.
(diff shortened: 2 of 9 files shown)


input/input.json
```diff
@@ -54,6 +54,11 @@
                             "School": {"alpha": 0.0},
                             "Community": {"alpha": 0.0}},
         "itinerary_ratios": {"Home": 0.25, "Work": 0.25, "School": 0.25, "Community": 0.25},
+        "weekends": {
+            "delay": 3.0,
+            "prop_school_time_to_home": 0.5,
+            "prop_school_time_to_comm": 0.5
+        },
```
confunguido: I would create a more general struct for "global modifiers" that are applicable for the entire duration of the simulation.

confunguido (changes requested):

src/school_calendar.rs
```diff
@@ -0,0 +1,144 @@
...
+    delay: f64,
+) -> ItineraryTransitionMatrix {
+    let weekend_matrix = [
+        [0.0, 0.0, 0.0, 0.0],
+        [0.0, 0.0, 0.0, 0.0],
+        [prop_school_time_to_home, 0.0, 0.0, prop_school_time_to_comm],
+        [0.0, 0.0, 0.0, 0.0],
+    ];
+    let acceptance: AcceptanceFunction = Rc::new(move |context, _person| {
+        context.get_current_time() % 7.0 >= delay && context.get_current_time() % 7.0 <= delay + 2.0
```
RobertJacobsonCDC: This is a subtle point, but do you want to include both left and right endpoints here?

src/itinerary_manager.rs
```diff
@@ -15,6 +15,7 @@ pub trait ItineraryModifierTrait: std::fmt::Debug + DynClone + 'static {
     fn layer(&mut self, other: Box<dyn ItineraryModifierTrait>) -> Box<dyn ItineraryModifierTrait>;
     fn apply(&mut self, base_itinerary: &[f64; SETTING_COUNT]) -> [f64; SETTING_COUNT];
     fn as_any(&self) -> &dyn Any;
+    fn accept(&self, context: &Context, person_id: PersonId) -> bool;
```
RobertJacobsonCDC, line 18: Consider giving `ItineraryModifierTrait::accept` a "default implementation" that always returns `true`.

RobertJacobsonCDC (commented):
## What needs _mutable_ reference to the `ItineraryModifierTrait`?

- `ItineraryModifierTrait::apply`, because we want to cache certain computations (but we haven't yet implemented this) — this is a good use case for "interior mutability" `OnceCell` for example. This let's you get rid of `&mut self` in favor of `&self`.
- `ItineraryModifierTrait::layer` is more complicated. We create a new `ItineraryModifierTrait` from `self` and `other: ItineraryModifierTrait`. Consider an alternative design that instead of mutating `self` we create a whole new `ItineraryModifierTrait`.


**Big picture:** What if we can get rid of all _mutation_ of instances of `ItineraryModifierTrait` once they are created and stored in the registry / storage. Then (consequences):

- They don't need to be clonable. You don't have to make copies. This gets rid of the dyn_clone.
- You don't need the `AcceptanceFunction` to be stored in an `Rc` (you can just store a function pointer / reference / `Box`).
- If you cache precomputations, you get to reuse them across _all_  uses of the itinerary modifiers, not just the copied versions you copied out of storage.


Known limitations:

- `ItineraryTransitionMatrix::absorption_probabilities` is being computed and cached but isn't being reused at the moment, just thrown away. 


## Editorial Comments

These are matters of subjective personal style, and they might be out of scope for this PR, so take them with a grain of salt:

- I avoid `Trait`, `Struct`, `Enum`, and similar suffixes on type names. They rarely add any value. So `ItineraryModifier` instead of `ItineraryModifierTrait`, `ItineraryModifierStorage` instead of `ItineraryModifierStorageTrait`, etc.

- I'd make `PersonPropertyItineraryModifier` into its own struct. This gives you a little more type safety for free. Your type alias includes a lifetime parameter `'a`, but it's not clear why it is needed. If you don't need it, get rid of it. Otherwise, attach it to where you need it. For example, if the `dyn ItineraryModifierTrait` holds a reference with lifetime `'a`, you can attach it to that parameter.

  ```rust
  // Without lifetime parameter
  struct PersonPropertyItineraryModifier<P> = (
      P,
      HashMap<P, Vec<Box<dyn ItineraryModifierTrait>>>,
  );
  
  // With lifetime attached to the trait object
  struct PersonPropertyItineraryModifier<'a, P>(
      P,
      HashMap<P, Vec<Box<dyn ItineraryModifierTrait + 'a>>>,
  );
  ```

src/itinerary_manager.rs
RobertJacobsonCDC: This might be out of scope for this PR, but I like to return the removed value (usually wrapped in an `Option`) for methods like this even if I don't use it at any of my existing call sites. That way the caller can take ownership if it wants, and if the caller doesn't want the removed value it just immediately gets dropped. It also gives the caller a mechanism to know if a value really was removed. As written, if `property_value` isn't actually a key in the underlying hash map, nothing actually happens, but the caller can't know that nothing happens. It's a really easy change, because `HashMap<K, V>::remove` already returns an `Option<V>` (for these same reasons).

src/itinerary_manager.rs
```diff
@@ -128,43 +131,62 @@ pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
...
             )
         {
-            let (_property, itinerary_modifier_map) = downcast_property_modifier_map;
+            let itinerary_modifier_map = &mut downcast_property_modifier_map.itinerary_modifier_map;
             itinerary_modifier_map.remove(&property_value);
         }
     }
 
-    fn get_itinerary_modifiers(&self, person_id: PersonId) -> Vec<Box<dyn ItineraryModifierTrait>>;
+    fn get_itinerary_modifiers(&self, person_id: PersonId) -> Vec<&dyn ItineraryModifier>;
```
RobertJacobsonCDC, line 130: It's a good habit to annotate getter methods with `#[must_use]`, because it's almost certainly an error to call such a function without actually doing something with the value it returns. It's a way of preemptively catching bugs. (But I admit I don't always practice what I preach.)

src/itinerary_manager.rs
```diff
@@ -9,51 +9,52 @@ use crate::{
...
-    ) -> Option<Vec<Box<dyn ItineraryModifierTrait>>>;
+    ) -> Option<Vec<&'a dyn ItineraryModifier>>;
     fn as_any(&self) -> &dyn Any;
     fn as_any_mut(&mut self) -> &mut dyn Any;
 }
 
-type PersonPropertyItineraryModifier<'a, P> = (
-    P,
+#[derive(Debug)]
+struct PersonPropertyItineraryModifier<P> {
```
RobertJacobsonCDC, line 39: The `ItineraryModifierStorage` type is the type-erased version, while the `PersonPropertyItineraryModifier<P>` type is the concrete type that knows the concrete property type `P`. Both of these are storage containers for a collection of `ItineraryModifiers`. In particular, the `PersonPropertyItineraryModifier<P>` type does not itself represent an itinerary modifier. So I would probably name it differently. Maybe `PersonPropertyItineraryModifierStorage<P>`, or `PropertyItineraryModifierStorage<P>`.

src/itinerary_manager.rs
```diff
@@ -9,51 +9,52 @@ use crate::{
...
 where
     P: IndexableProperty<Person>,
 {
     fn get_itinerary_modifiers(
         &self,
         context: &Context,
         person_id: PersonId,
-    ) -> Option<Vec<Box<dyn ItineraryModifierTrait>>> {
-        let (_person_property, modifier_map) = self;
+    ) -> Option<Vec<&dyn ItineraryModifier>> {
```
RobertJacobsonCDC: This is probably fine as is, but I just want to highlight an alternative: Instead of returning an `Option<Vec>`, you can just return a `Vec`. If there are no itinerary modifiers for the given person, the `Vec` would just be empty. Creating an empty `Vec` is guaranteed not to allocate, so it's really efficient to use `Vec` in this way.

You would still want an `Option<Vec>` if there is a real semantic difference between `self.itinerary_modifier_map` not having `property_val` as a key (the `None` case) versus the key existing with an empty value.

src/itinerary_manager.rs
```diff
@@ -92,20 +93,22 @@ pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
                 .as_any_mut()
                 .downcast_mut::<PersonPropertyItineraryModifier<P>>(
             ) {
-                let (_property, itinerary_modifier_map) = downcast_modifier_map;
-                itinerary_modifier_map
+                let downcast_modifier_map: &mut PersonPropertyItineraryModifier<P> =
+                    downcast_modifier_map;
```
RobertJacobsonCDC: I don't think this `let` statement is actually doing anything. I think this line can be deleted.

src/itinerary_manager.rs
```diff
@@ -78,7 +79,7 @@ define_data_plugin!(
 
 pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
     /// Register a generic itinerary modifier.
-    fn register_itinerary_modifier<P: IndexableProperty<Person>, I: ItineraryModifierTrait>(
+    fn register_itinerary_modifier<P: IndexableProperty<Person>, I: ItineraryModifier>(
```
RobertJacobsonCDC, line 84: You should check me on this, but I think the following implementation of this method is semantically equivalent and a little easier to read and understand:

```rust
fn register_itinerary_modifier<P, I>(
    &mut self,
    person_property: P,
    itinerary_modifier: I,
) where
    P: IndexableProperty<Person>,
    I: ItineraryModifier,
{
    let storage = self
        .get_data_mut(ItineraryModifierPlugin)
        .itinerary_modifier_map
        .entry(TypeId::of::<P>())
        .or_insert_with(|| {
            Box::new(PersonPropertyItineraryModifier::<P> {
                itinerary_modifier_map: HashMap::new(),
            })
        });

    let modifier_map = storage
        .as_any_mut()
        .downcast_mut::<PersonPropertyItineraryModifier<P>>()
        .expect("itinerary modifier storage has the wrong type");

    modifier_map
        .itinerary_modifier_map
        .entry(person_property)
        .or_default()
        .push(Box::new(itinerary_modifier));
}
```

You can probably safely replace the `expect` with an `unwrap`, because the key `TypeId::of::<P>()` is guaranteed to have a value of the correct type, and so the downcast should be infallible. But this is a judgment call. If the downcast fails, you probably want the program to crash, because that indicates a serious bug in your implementation. In your original version, the registration just silently fails, which I don't think it what you want.

src/itinerary_manager.rs
```diff
@@ -92,20 +93,22 @@ pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
                 .as_any_mut()
                 .downcast_mut::<PersonPropertyItineraryModifier<P>>(
```
RobertJacobsonCDC: If this downcast fails, your registration silently fails, which probably isn't what you want. I'd use either `expect` or `unwrap` instead of an `if let` construct in this case so that an internal bug that causes the downcast to fail just crashes the program.

src/itinerary_manager.rs
```diff
@@ -128,43 +131,62 @@ pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
...
+
+        container
+            .itinerary_modifier_map
+            .values()
+            .filter_map(|modifier_map| modifier_map.get_itinerary_modifiers(self, person_id))
+            .flatten()
+            .collect()
     }
 
     fn get_itinerary(&self, person_id: PersonId) -> [f64; SETTING_COUNT] {
```
RobertJacobsonCDC, line 146: This one is a little tricky because of both the control flow and the types involved. Again, check me on this, but I think the following is semantically equivalent but easier to read and understand:

```rust
fn get_itinerary(&self, person_id: PersonId) -> [f64; SETTING_COUNT] {
    let base_itinerary = self
        .get_property::<Person, Itinerary>(person_id)
        .itinerary_ratios;

    let mut modifiers = self
        .get_itinerary_modifiers(person_id)
        .into_iter()
        .filter(|modifier| modifier.accept(self, person_id));

    let Some(first) = modifiers.next() else {
        return base_itinerary;
    };

    let Some(second) = modifiers.next() else {
        return first.apply(&base_itinerary);
    };

    modifiers
        .fold(first.layer(second), |layered, modifier| {
            layered.layer(modifier)
        })
        .apply(&base_itinerary)
}
```

The reason we pull out the first _two_ modifiers is,  if you only have zero or one modifier, you don't need to construct a boxed modifier, but if you have two or more, that first `layer` construction gives you the correct type (`Box<dyn ItineraryModifier>` instead of `&dyn ItineraryModifier`).

As an aside, there are actually two really good reasons to prefer the iterator combinator methods like `filter`, `fold`, `map`, etc.:

1. They often make the code much easier to read and understand, and
2. the compiler can often produce extremely optimized machine code when you use them, often better than hand written loops and control flow.

src/itinerary_manager.rs
```diff
@@ -128,43 +131,62 @@ pub trait ContextItineraryModifierExt: PluginContext + ContextEntitiesExt {
                 .downcast_mut::<PersonPropertyItineraryModifier<P>>(
```
RobertJacobsonCDC: Similar to my comment on `register_itinerary_modifier` on line 94 above: If this downcast fails, the method silently fails, which is probably not what you want. Instead you can separate the destructing of the `modifier_map` from the downcast:

```rust
if let Some(property_modifier_map) = modifier_map {
    let downcast = property_modifier_map
        .as_any_mut()
        .downcast_mut::<PersonPropertyItineraryModifier<P>>()
        .expect("modifier map entry had unexpected type for its TypeId key");
    downcast.itinerary_modifier_map.remove(&property_value);
}
```

src/school_calendar.rs
```diff
@@ -0,0 +1,344 @@
...
+pub struct SchoolCalendarModifier {
+    pub modifier: SchoolCalendarModifierType,
+    pub start_time: f64,
+    pub prop_school_time_to_home: f64,
+    pub prop_school_time_to_comm: f64,
+    pub end_time: Option<f64>,
+}
+
+impl SchoolCalendarModifier {
+    pub fn validate(&self) -> Result<(), ModelError> {
```
RobertJacobsonCDC, line 37: I would make it impossible to construct an invalid `SchoolCalendarModifier` by making the fields private and putting the validation code in a public constructor method.

src/itinerary_modifiers.rs
RobertJacobsonCDC, line 153: You might consider factoring this out into a free function `fn invert<const N: usize>(mut matrix: [[f64; N]; N]) -> Option<[[f64; N]; N]>`. In the implementation, you can return `None` if you detect singularity:

```rust
fn invert<const N: usize>(
    mut matrix: [[f64; N]; N],
) -> Option<[[f64; N]; N]> {
    let mut inverse = [[0.0; N]; N];

    for (i, row) in inverse.iter_mut().enumerate() {
        row[i] = 1.0;
    }

    for col in 0..N {
        let pivot_row = (col..N)
            .max_by(|&left, &right| {
                matrix[left][col]
                    .abs()
                    .total_cmp(&matrix[right][col].abs())
            })
            .expect("pivot range is nonempty");

        if matrix[pivot_row][col].abs() <= f64::EPSILON {
            return None;
        }

        matrix.swap(col, pivot_row);
        inverse.swap(col, pivot_row);

        let pivot = matrix[col][col];
        for j in 0..N {
            matrix[col][j] /= pivot;
            inverse[col][j] /= pivot;
        }

        for row in 0..N {
            if row == col {
                continue;
            }

            let factor = matrix[row][col];
            for j in 0..N {
                matrix[row][j] -= factor * matrix[col][j];
                inverse[row][j] -= factor * inverse[col][j];
            }
        }
    }

    Some(inverse)
}
```

src/school_calendar.rs
RobertJacobsonCDC: This also makes it impossible to forget to validate a `SchoolCalendarModifier`.

src/school_calendar.rs
```diff
@@ -0,0 +1,344 @@
...
+            0.0,
+            0.0,
+            params.prop_school_time_to_comm,
+        ],
+        [0.0, 0.0, 0.0, 0.0],
+    ];
+
+    match params.modifier {
+        SchoolCalendarModifierType::Weekend => {
+            let start_time = params.start_time;
```
confunguido: I did not understand that this means the start of the weekend. Be more specific.

src/school_calendar.rs
```diff
@@ -0,0 +1,344 @@
...
+                context.get_current_time() % 7.0 >= start_time
+                    && context.get_current_time() % 7.0 < start_time + 2.0
+            });
+            Ok(create_itinerary_transition_matrix(
+                Some(matrix),
+                None,
+                Some(acceptance),
+            ))
+        }
+        _ => {
```
confunguido, line 124: THis needs to repeat every year.

